PR #1, opened 2019-11-04
Add synchronization script

i386x:
Add a script that propagates files that should be in sync across linux-system-roles repos. The propagating is done via pull request.


sync-template.sh
```diff
@@ -0,0 +1,424 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+
+set -euo pipefail
+
+ME=$(basename $0)
+HERE=${PWD}
+WORKDIR="${HOME}/.cache/${ME%.*}"
+STDERR=/tmp/${ME%.*}.stderr
```
tyll: Files in `/tmp` need to be created with mktemp for security reasons, see for example: https://en.wikipedia.org/wiki/Symlink_race

richm:
any updates?

i386x:
@richm, the script is almost complete now. Things that should be done are:
- move `--clean` option out (provide it as a separate script instead)
- write a documentation (update `README.md`, provide some examples)

Things that should be done in future (not in this PR):
- provide scripts for repo health checking (like if there are no obsolete template files, conflicting branches, etc.)
- prevent copying of `lgtm.yml` if there are no Python files
- make a `tox` plugin that can help to reduce a large number of shell wrappers and the size of `tox.ini` itself

i386x:
Also, a little cleanup in commit history (some commits, like license change, shoud be moved to separate pull request) before merging this will be useful.

richm:
@i386x is this PR ready to merge?  Was there additional work required?

i386x:
Hi Rich, I am planning to squash some commits to make a history more clear and also move some unrelated commits (like `Change license to MIT`) to separate pull requests.

i386x:
I think this can be merged.

richm:
> I think this can be merged.

Does this PR need to be rebased and squashed?  Or do we want all of those commits listed separately?

PR #6, opened 2020-03-24
ability to run a command in each repo

richm:
Useful for a linux-system-roles "administrator" role where you
want to have a local, up-to-date clone of all role repos, and
possibly run some sort of arbitrary command against all of them,
and a special purpose script like `sync-template.sh` isn't
suitable.  Also creates an upstream fork in your personal repo
if you don't already have one (so you can do feature work, submit
pull requests, etc.).
Requires `hub` and `jq`

### DIFF
--- a/local-repo-dev-sync.sh
+++ b/local-repo-dev-sync.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
+
+set -euo pipefail
+
+if [ "${DEBUG:-false}" = true ]; then
+    set -x
+fi
+
+LSR_GH_ORG=${LSR_GH_ORG:-linux-system-roles}
+LSR_BASE_DIR=${LSR_BASE_DIR:-~/linux-system-roles}
+
+requiredcmds="hub jq"
+missingcmds=""
+
+for cmd in $requiredcmds; do
+    if ! type -p "$cmd" > /dev/null 2>&1; then
+        missingcmds="$missingcmds $cmd"
+    fi
+done
+
+if [ -n "$missingcmds" ]; then
+    echo ERROR: this script requires the following commands: "$missingcmds"
+    echo e.g. on Fedora - dnf -y install "$missingcmds"
+    exit 1
+fi
+
+if [ ! -d "$LSR_BASE_DIR" ]; then
+    mkdir -p "$LSR_BASE_DIR"
+fi
+
+##
+# GET (list) a given github url, using the page interface to get all
+# of the pages, and using a given jq filter to extract fields
+#
+#    $1 - github url e.g. orgs/linux-system-roles/repos
+#    $2 - jq filter e.g. '.[].name'
+#
+# echos to stdout the filtered results.
+# NOTE: the -e argument to jq tells jq to exit with an error code
+# if the input is empty - this breaks the loop
+gh_get_all() {
+    local uri="$1"
+    local filter="$2"
+    local page=1
+    while hub api "${uri}?page=$page" | jq -e -r "$filter"; do
+        page=$((page + 1))
+    done
+}
+
+repos=${REPOS:-$(gh_get_all orgs/linux-system-roles/repos '.[].name')}
+EXCLIST=${EXCLIST:-"test-harness linux-system-roles.github.io sap-base-settings \
+                     sap-hana-preconfigure experimental-azure-firstboot sap-preconfigure \
+                     auto-maintenance image_builder sap-netweaver-preconfigure ci-testing \
+                     meta_test tox-lsr tuned postfix"}
+declare -A EXARRAY
+for repo in $EXCLIST; do
+    # EXARRAY is a "set" of excluded repos
+    EXARRAY[$repo]=$repo
+done
+
+pushd "$LSR_BASE_DIR" > /dev/null
+if ! tty -s; then
+    stdincmds="$(cat)"
+fi
+
+for repo in $repos; do
+    if [ -n "${EXARRAY[$repo]:-}" ]; then
+        continue
+    fi
+
+    echo Repo: "$repo"
+    # get a local clone of the repo
+    if [ ! -d "$LSR_BASE_DIR/$repo" ]; then
+        HUB_PROTOCOL=https hub clone "$LSR_GH_ORG/$repo"
+    fi
+    pushd "$LSR_BASE_DIR/$repo" > /dev/null
+    # should have a remote called origin that points to lsr/repo
+    if ! git remote get-url origin | grep -q "$LSR_GH_ORG/$repo"; then
+        echo Error: non-standard git remote config - origin does not point
+        echo to "$LSR_GH_ORG/$repo"
+        git remote get-url origin
+        echo please use git remote to configure origin to point to "$LSR_GH_ORG/$repo"
+        exit 1
+    fi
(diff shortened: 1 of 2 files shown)


local-repo-dev-sync.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo see https://stedolan.github.io/jq/manual/
+    echo e.g. on Fedora - dnf -y install jq
+    exit 1
+fi
+
+if [ ! -d $LSR_BASE_DIR ] ; then
+    mkdir -p $LSR_BASE_DIR
+fi
+
+pushd $LSR_BASE_DIR > /dev/null
```
tyll, line 33: There is no popd, is this intentional?

local-repo-dev-sync.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo Repo: $repo
+    if [ $repo = kernel_settings ] ; then
+        echo skipping
+        continue
+    fi
+    # get a local clone of the repo
+    if [ ! -d $LSR_BASE_DIR/$repo ] ; then
+        HUB_PROTOCOL=https hub clone $LSR_GH_ORG/$repo
+    fi
+    cd $LSR_BASE_DIR/$repo
```
tyll, line 58: Why `cd` instead of pushd here?

local-repo-dev-sync.sh
```diff
@@ -0,0 +1,78 @@
...
+    # this will also create a git remote in the local repo if there
+    # is not already one
+    HUB_PROTOCOL=ssh hub fork
+    git fetch
+    if [ -n "${1:-}" ] ; then
+        if ! "$@" ; then
+            echo ERROR: command in $( pwd ) failed
+        fi
+    fi
+    cd ..
```
tyll, line 77: This should be `popd´ when using `pushd` instead of `cd`. Also `cd -` is probably more correct as it returns to $OLDPWD AFAIU.

local-repo-dev-sync.sh
richm: added a popd

local-repo-dev-sync.sh
richm: changed to pushd

mprovenc:
test

richm:
@spetrosi

local-repo-dev-sync.sh
```diff
@@ -0,0 +1,109 @@
...
+#    $2 - jq filter e.g. '.[].name'
+#
+# echos to stdout the filtered results.
+# NOTE: the -e argument to jq tells jq to exit with an error code
+# if the input is empty - this breaks the loop
+gh_get_all() {
+    local uri="$1"
+    local filter="$2"
+    local page=1
+    while hub api "${uri}?page=$page" | jq -e -r "$filter"; do
```
spetrosi, line 46: For this to work, I needed to create a Personal access token in GitHub and set the `GITHUB_TOKEN` environment variable in my OS to point to the token. Without these steps, the script hangs at this line without any error message.

spetrosi: When I've set the GITHUB_TOKEN var, the script worked fine.

richm: hmm - also, hub uses `~/.config/hub`:
```
github.com:
- user: richm
  oauth_token: xxxxxxxxxxxxxx
  protocol: https
```
I thought there was some hub command like hub login to create the initial token and write the file - It's been a while since I did the hub setup - maybe on the hub man page it specifies this

local-repo-dev-sync.sh
```diff
@@ -0,0 +1,109 @@
...
+    if ! type -p "$cmd" > /dev/null 2>&1; then
+        missingcmds="$missingcmds $cmd"
+    fi
+done
+
+if [ -n "$missingcmds" ]; then
+    echo ERROR: this script requires the following commands: "$missingcmds"
+    echo e.g. on Fedora - dnf -y install "$missingcmds"
+    exit 1
+fi
```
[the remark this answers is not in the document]
richm: hmm - when I try to run on a system that does not have `hub`:
```
ERROR: this script requires the following commands:  hub
e.g. on Fedora - dnf -y install  hub
```

local-repo-dev-sync.sh
richm: from the `hub` man page:
```
   GitHub OAuth authentication
       Hub will prompt for GitHub username  password the first time it needs
       to access the API and exchange it for an OAuth token, which it saves in
       ~/.config/hub.
```
So I guess just run something like `hub api /repos/linux-system-roles/mssql` and it should prompt for username and password, and create `~/.config/hub` for you

local-repo-dev-sync.sh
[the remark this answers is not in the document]
spetrosi: Yeah, no problem here, I am just confirming that this works.

local-repo-dev-sync.sh
spetrosi: When I provide my credentials to `hub api`, it return an error saying `Not Found`:
```
$ hub api "orgs/linux-system-roles/repos?page=1" 
github.com username: spetrosi@redhat.com
github.com password for spetrosi@redhat.com (never stored): 
Not Found
```
I guess that this might happen because I have two-factor auth configured in GitHub, can this be the case? I only managed to configure this by creating a token and setting the env var.

richm: > When I provide my credentials to hub api, it return an error saying Not Found:

I suppose it could be related to 2FA, but I would expect a 401 or 403 instead of Not Found - you could use `hub api -i ...` and see if you can determine what is the actual error from the headers returned.

But at any rate - if you are ok with setting the token environment variable every time, or if you create a `~/.config/hub` with your token

spetrosi: I am completely ok with doing this, but the script must say that there is such a requirement. Otherwise, there is a gap because now the script only asks to install hub.

local-repo-dev-sync.sh
```diff
@@ -25,6 +25,16 @@ if [ -n "$missingcmds" ]; then
...
+if [ -n "$GITHUB_TOKEN" ] || [ -f "$HOME/.config/hub" ]; then
+    : # ok
+else
+    echo ERROR: no github credentials specified for the hub command.
+    echo Please see the manpage for the hub command for how to specify
+    echo credentials in your \$HOME/.config/hub config file.
+    echo Alternately, you can use the GITHUB_TOKEN environment variable.
+    exit 1
+fi
+
```
richm, line 37: @spetrosi try this - note that this will still fail if you have `$HOME/.config/hub` but it has a problem like wrong format, incorrect token, etc.

nhosoi:
@richm, it might be a bit tricky, but if you could put some usages / examples (as the next one you gave me) to this script or readme, that'd be really helpful. Other than that, you have my lgtm.
```
LSR_BASE_DIR=~/tmp/working-lsr/ DEBUG=true ./local-repo-dev-sync.sh << EOF
set -euxo pipefail
git checkout main || git checkout master
git checkout -b tox-lsr-2.2.0
sed -i -e 's|TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.1.2"|TOX_LSR: "git+https://github.com/linux-system-roles/tox-lsr@2.2.0"|' .github/workflows/tox.yml 
git commit -a -F ./commit.txt
EOF
```

PR #7, opened 2020-03-24
add copy=tests/setup_module_utils.sh

richm:


pcahyna:
thanks, I am trying it to synchronize https://github.com/linux-system-roles/template/pull/14 into https://github.com/linux-system-roles/timesync/pull/65

pcahyna:
Recent or propose changes to rempalte need more additions, which can be done here or in a new PR.

PR #8, opened 2020-03-29
add --local; add yamllint

richm:
Add a `--local` switch - this allows you to copy files from your
local clone in order to e.g. test with `tox` locally.
Add support for the yamllint files.

### DIFF
--- a/sync-template.sh
+++ b/sync-template.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 
 ME=$(basename $0)
 HERE=${PWD}
-WORKDIR="${HOME}/.cache/${ME%.*}"
+WORKDIR="${WORKDIR:-${HOME}/.cache/${ME%.*}}"
 STDOUT=$(mktemp /tmp/${ME%.*}-XXXXXX.stdout)
 STDERR=$(mktemp /tmp/${ME%.*}-XXXXXX.stderr)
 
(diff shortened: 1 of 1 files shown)


sync-template.sh
```diff
@@ -51,6 +51,8 @@ FILES=(
   '--copy-if-missing=pylint_extra_requirements.txt'
   '--copy=pylintrc'
   '--copy=tox.ini'
+  '--copy=.yamllint.yml'
+  '--copy-if-missing=.yamllint_custom.yml'
```
pcahyna: By the way, I made a PR for this - https://github.com/richm/auto-maintenance/pull/1 - maybe I should have advertized it more loud.

sync-template.sh
```diff
@@ -502,14 +522,19 @@ function do_sync() {
   )
 
   for REPO in ${REPOLIST}; do
-    inform "Synchronizing ${REPO} wiht ../${LSR_TEMPLATE}."
+    inform "Synchronizing ${REPO} with ../${LSR_TEMPLATE}."
     runcmd "[[ -d \"${REPO}\" ]] && rm -rfd ${REPO} || :"
     runcmd "git clone '${GITHUB}/${LSR_GROUP}/${REPO}.git' '${REPO}'"
```
pcahyna: Does this work with --local ?

sync-template.sh
```diff
@@ -502,14 +522,19 @@ function do_sync() {
   )
 
   for REPO in ${REPOLIST}; do
-    inform "Synchronizing ${REPO} wiht ../${LSR_TEMPLATE}."
+    inform "Synchronizing ${REPO} with ../${LSR_TEMPLATE}."
     runcmd "[[ -d \"${REPO}\" ]] && rm -rfd ${REPO} || :"
```
pcahyna: Isn't this dangerous with --local ? IIUC the idea of --local is that WORKDIR is not being treated as a temporary directory.

sync-template.sh
richm: no, it doesn't - fixed

sync-template.sh
richm: yes, fixed

sync-template.sh
richm: Yeah, not sure why I didn't see this earlier

pcahyna: If you accept this suggestion https://github.com/linux-system-roles/network/pull/178#discussion_r399975678 it will need to change.

PR #9, opened 2020-04-02
allow the use of the script with the hub command

richm:
Use the `hub` command to allow workflows that use your
local git user with your local git credentials e.g.

As a regular user, perform the usual
clone repo as origin
fork repo as myusername
make a change locally
push the change to my forked repo to a branch
submit a pull request from my forked repo to origin

This also makes the sync script both idempotent and able
to run incrementally.  That is, you can have the following
workflow:

make changes to template
sync the changes to repo R branch B
create a PR from branch B
make more changes to template
sync the changes to repo R branch B
push changes to branch B

This will update the existing PR with a new commit with the
additional changes on branch B

This PR also addresses a major limitation of the sync script -
it does not add new files to git.

### DIFF
--- a/sync-template.sh
+++ b/sync-template.sh
@@ -26,8 +26,13 @@ LSR_GROUP="linux-system-roles"
 LSR_TEMPLATE="template"
 LSR_TEMPLATE_NS="${LSR_GROUP}/${LSR_TEMPLATE}"
 LSR_TEMPLATE_REPO="${GITHUB}/${LSR_TEMPLATE_NS}.git"
+LSR_FORK_PREFIX=${LSR_FORK_PREFIX:-"${LSR_GROUP}-"}
+# this is the name of the git remote to use for the user's
+# fork of the lsr repo
+LSR_FORK_REMOTE=${LSR_FORK_REMOTE:-"lsr-user-remote"}
 
 FILES=(
+  '--copy=.ansible-lint'
   '--copy-if-missing=.gitignore'
   '--copy-if-missing=.lgtm.yml'
   '--ensure-directory=.travis'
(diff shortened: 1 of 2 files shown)


richm:
OK - I think I have all of the kinks fixed, with odd corner cases (git remote exists but not actual remote repo; user already has user/projectA and attempts to fork linux-system-roles/projectA, etc.)

As an administrator of linux-system-roles, I find the workflow made possible by these changes extremely useful.  I'm would like to merge this, even if no one else finds it currently useful.

BTW, if you haven't used `hub` for interactive github sessions, I highly recommend it.

sync-template.sh
```diff
@@ -65,6 +71,29 @@ INHELP=""
...
 
+if type -p hub > /dev/null 2>&1 ; then
+  if [[ -z "${USE_HUB:-}" ]]; then
+    USE_HUB=true
+  fi
+elif [[ "${USE_HUB:-false}" = true ]]; then
+  echo ERROR: there is no \"hub\" command line tool
+  echo see https://github.com/github/hub
+  echo e.g. on Fedora - dnf -y install hub
+  exit 1
```
tyll, line 82: Small nitpick: It would be nicer for users to check all dependencies first and then fail with a list of all missing dependencies instead of checking each dependency individually. Then the user only needs to install the packages once instead of switching between running the script and install dependencies.

sync-template.sh
```diff
@@ -429,6 +501,25 @@ CONTACTS="${CONTACTS:-${CONTACTS_DEFAULT}}"
 export GITHUB_TOKEN="${GITHUB_TOKEN:-}"
 REPOLIST="${REPOLIST:-}"
 REPOLIST="${REPOLIST//,/ }"
+if [[ $USE_HUB = true ]]; then
+  case $FROM_REPO in
+      http*://github.com/*) FROM_REPO_HUB=${FROM_REPO#http*://github.com/} ;;
```
tyll, line 506: Does it need to support plain `http://` - I hope everything is using https and if not it should be fixed.

richm: the use of `http://` is now an error

pcahyna:
> As an administrator of linux-system-roles, I find the workflow made possible by these changes extremely useful. I would like to merge this, even if no one else finds it currently useful.

Is there documentation how to use the script with this workflow, please? Above, you outlined the workflow, but not the actual commands / options to perform the individual steps.

richm:
> > As an administrator of linux-system-roles, I find the workflow made possible by these changes extremely useful. I would like to merge this, even if no one else finds it currently useful.
> 
> Is there documentation how to use the script with this workflow, please? Above, you outlined the workflow, but not the actual commands / options to perform the individual steps.

I've added usage and workflow descriptions to the README.md in the latest commit.

richm:
I'd like to merge this.  Any objections?

PR #11, opened 2020-05-06
fix automated mode

richm:
My previous commit broke automated mode.  This fixes automated mode
and cleans up the code a bit.  There are two new cmdline flags:

`--workdir /path/to/dir` - instead of using the `$WORKDIR` env. var.,
pass it in with this flag.  This will automatically set `PRESERVE=true`
so as not to accidentally remove any directories under your workdir.

`--use-hub` - previously it would use `hub` automatically if found
in the `$PATH`.  Now, you must explicitly set `--use-hub` in order
to use `hub`.


richm:
This should fix the issues found during the demo, and address the questions @pcahyna asked.

PR #12, opened 2020-05-06
shellcheck

richm:


richm:
@robled

sync-template.sh
```diff
@@ -69,6 +69,7 @@ declare -A IGNORE_IF_MISSING_MAP
 INDENT=""
 INHELP=""
 
+# shellcheck disable=SC2064
```
tyll, line 72: Does shellcheck support names instead of codes? If you add a name instead of a code, it would be easier to understand what is disabled here.

richm: > Does shellcheck support names instead of codes? If you add a name instead of a code, it would be easier to understand what is disabled here.

Looks like no names - codes only.  However, each code has a separate wiki page e.g. https://github.com/koalaman/shellcheck/wiki/SC2016

If this is a problem, I suppose we could have additional comments for each supression e.g.
```
# https://github.com/koalaman/shellcheck/wiki/SC2016
# Expressions don't expand in single quotes, use double quotes for that.
# shellcheck disable=SC2064 
xx='$i_dont_want_to_expand_the_var_here'
```

On the other hand, we use the same code-only style with other linters used in system roles e.g. for a pylint supression
```
        raise caught_import_error  # pylint: disable-msg=E0702
```

tyll: I don't use codes in the network role but names like `# pylint: disable=import-error, no-name-in-module` ;-). Do, what you like best.

richm: My intention is that `shellcheck` will become one of the standard `tox` CI tests for checking shell scripts in all of the repos, including `network`, so you should indicate your preference.

tyll: I filed https://github.com/koalaman/shellcheck/issues/1948 to request name IDs for checks.

My preference is to add at least a reference to the wiki. The text `Expressions don't expand in single quotes, use double quotes for that.` seems to be irritating for me in the commend since it is not clear at first sight that this is the behavior that shellcheck wants. So something like `ignore expansions in single quotes` would be better in this case.

pcahyna:
This does not run shellcheck on the roles, just on the auto_maintenance repository itself, right?

richm:
> This does not run shellcheck on the roles, just on the auto_maintenance repository itself, right?

correct - we will need to create a `tox` test in `template`, then sync it to the roles

sync-template.sh
richm: see https://github.com/linux-system-roles/auto-maintenance/pull/12/commits/d999ca3f5a10bc9e46c3376877593ca920276ac7
OK?

richm:
> > This does not run shellcheck on the roles, just on the auto_maintenance repository itself, right?

> correct - we will need to create a `tox` test in `template`, then sync it to the roles

shellcheck has been added to the template.  Before making the changes to `sync-template.sh` to allow it to sync those changes to the roles, I would like to merge this PR.  @tyll had some concerns - I think they have been addressed, and I refactored this PR as suggested by @tyll in another PR to have the first commit to add shellcheck to the CI, and a subsequent commit to address the issues found by shellcheck.

richm:
ok to merge?

richm:
ok to merge?

richm:
ok to merge?

PR #26, opened 2020-10-27
lsr_role2collection.py - trailing comments were not properly handled

nhosoi:
Import statement with a trailing comment like this
  from ansible.module_utils.network_lsr import MyError  # noqa:E501
has to be converted to
  from ansible_collections.fedora.system_roles.plugins.module_utils.network_lsr.__init__ import MyError  # noqa:E501

While examinig the path in module_utils, the comment had to be ingnored
then put back in the converted line.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -381,7 +381,7 @@ DO_NOT_COPY = (
 ALL_DIRS = ROLE_DIRS + PLUGINS + TESTS + DOCS + DO_NOT_COPY
 
 IMPORT_RE = re.compile(
-    br"(\bimport) (ansible\.module_utils\.)(\S+)(\S*)(\s+#.+|.*)$", flags=re.M
+    br"(\bimport) (ansible\.module_utils\.)(\S+)(.*)(\s+#.+|.*)$", flags=re.M
 )
 FROM_RE = re.compile(
     br"(\bfrom) (ansible\.module_utils\.?)(\S+)? import (\(*(?:\n|\r\n)?)(\S+)(\s+#.+|.*)$",
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -380,9 +380,11 @@ DO_NOT_COPY = (
 
 ALL_DIRS = ROLE_DIRS + PLUGINS + TESTS + DOCS + DO_NOT_COPY
 
-IMPORT_RE = re.compile(br"(\bimport) (ansible\.module_utils\.)(\S+)(.*)$", flags=re.M)
+IMPORT_RE = re.compile(
+    br"(\bimport) (ansible\.module_utils\.)(\S+)(\S*)(\s+#.+|.*)$", flags=re.M
+)
 FROM_RE = re.compile(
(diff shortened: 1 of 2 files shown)


lsr_role2collection.py
```diff
@@ -380,9 +380,11 @@ def run(self):
 
 ALL_DIRS = ROLE_DIRS + PLUGINS + TESTS + DOCS + DO_NOT_COPY
 
-IMPORT_RE = re.compile(br"(\bimport) (ansible\.module_utils\.)(\S+)(.*)$", flags=re.M)
+IMPORT_RE = re.compile(
+    br"(\bimport) (ansible\.module_utils\.)(\S+)(\S*)(\s+#.+|.*)$", flags=re.M
```
richm, line 384: how does the regexp parser know the difference/boundary between `(\S+)(\S*)` since they both match almost the same thing?

nhosoi: Well, if there's no a trailing comment, the last match (match.group(5)) is supposed to be empty. I.e., what I want to have from `(\s+#.+|.*)` is if the line ends with `  # noqa:E501`, set it to match.group(5) otherwise none. Do you think `(\s+#.+|.*)` is a reasonable regex for the purpose?

richm: Can you show me an example of something that `(\S+)(\S*)` matches, and something that it doesn't match?

nhosoi: You are right. I found I had no example which matches such a case. Rather, it turned out we were just lucky not to have, e.g., `import ansible.module_utils.util0, ansible.module_utils.util1, ansible.module_utils.util2, ...`
Let me rework on this pr. I'm setting WIP for now.

nhosoi: @richm, fixed `(\S+)(\S*)`, which should have been `(\S+)(.*)` to capture as localname. And added the test cases.
```
    Adding test cases
      import ansible.module_utils.utils0  # noqa:E501
      import ansible.module_utils.utils1 as local1  # noqa:E501
    matching
      "(\bimport) (ansible\.module_utils\.)(\S+)(.*)(\s+#.+|.*)$"
    to test_lsr_role2collection.py.
```
Regarding `import ansible.module_utils.utils0, ansible.module_utils.utils1, ansible.module_utils.utils2`, I'd like to propose to give it up in the system roles. (See also https://issues.redhat.com/browse/RHELPLAN-54079?focusedCommentId=15375157&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15375157)

richm: I think tools like black, flake8, pylint will reject `import ansible.module_utils.utils0, ansible.module_utils.utils1, ansible.module_utils.utils2` - if so, we definitely don't have to worry about that case - at any rate, I think we should not handle it

nhosoi: You are right! Flake8 complains it:
```
import errno, functools, os, re, shlex, socket, subprocess, time, traceback
            ^
1     E401 multiple imports on one line
```
We are good. :)

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #27, opened 2020-10-28
lsr_role2collection.py - converting the value of roletoinclude

nhosoi:
Converting
roletoinclude: linux-system-roles.ROLENAME
to
roletoinclude: fedora.system_roles.ROLENAME

It is a hack to support tests/tests_include_vars_from_parent.yml
in certificate, kernel_settings, logging, nbde_client, nbde_server,
storage, and vpn.

A side note: Do we want to add tests_include_vars_from_parent.yml and its helper code to the newer roles, e.g., cockpit? (I don't see vars in firewall.)

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -432,6 +432,10 @@ class LSRFileTransformer(LSRFileTransformerBase):
         """handle vars of Ansible item, or vars from a vars file"""
         for var in a_item.get("vars", []):
             logging.debug(f"\tvar = {var}")
+            if var == "roletoinclude":
+                lsr_rolename = self.src_owner + "." + self.rolename
+                if a_item["vars"][var] == lsr_rolename:
+                    ru_item["vars"][var] = self.prefix + self.rolename
         return
 
     def meta_cb(self, a_item, ru_item):


lsr_role2collection.py
```diff
@@ -432,6 +432,10 @@ def vars_cb(self, a_item, ru_item):
         """handle vars of Ansible item, or vars from a vars file"""
         for var in a_item.get("vars", []):
             logging.debug(f"\tvar = {var}")
+            if f"{var}" == "roletoinclude":
```
richm, line 435: do you need to convert `var` to a string?  If not then you could just do
```python
if var == "roletoinclude":
```

lsr_role2collection.py
```diff
@@ -432,7 +432,7 @@ def vars_cb(self, a_item, ru_item):
         """handle vars of Ansible item, or vars from a vars file"""
         for var in a_item.get("vars", []):
             logging.debug(f"\tvar = {var}")
-            if f"{var}" == "roletoinclude":
+            if var == "roletoinclude":
                 lsr_rolename = self.src_owner + "." + self.rolename
                 if a_item["vars"][f"{var}"] == lsr_rolename:
                     ru_item["vars"][f"{var}"] = self.prefix + self.rolename
```
richm, line 438: these two lines also - `f"{var}"` -> `var`

nhosoi: Sorry, I should have tried them... (;_;) Yes, no need to convert to a string there, too.
Replaced and squashed.

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #28, opened 2020-10-29
lsr_role2collection.py - the source role path does not have to end with the role name

nhosoi:
Instead of checking the existence of the path with the role name, use meta/main.yml.
    
Plus, instead of using print, use the logging module to output the log messages.
And introduced an env var LSR_INFO to output the info level messages.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -390,6 +390,8 @@ FROM_RE = re.compile(
 
 if os.environ.get("LSR_DEBUG") == "true":
     logging.getLogger().setLevel(logging.DEBUG)
+elif os.environ.get("LSR_INFO") == "true":
+    logging.getLogger().setLevel(logging.INFO)
 else:
     logging.getLogger().setLevel(logging.ERROR)
 
@@ -584,7 +586,7 @@ def copy_tree_with_replace(
                 dest = dest_path / "roles" / role / dirname
             else:
                 dest = dest_path / dirname / role
-            print(f"Copying role {src} to {dest}")
+            logging.info(f"Copying role {src} to {dest}")
             if ignoreme:
                 lsr_copytree(
                     src,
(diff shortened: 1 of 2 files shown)


lsr_role2collection.py
```diff
@@ -963,13 +965,19 @@ def role2collection():
...
-    _src_path = args.src_path.resolve()
+    src_path = args.src_path.resolve()
     src_owner = args.src_owner
     if not src_owner:
-        src_owner = os.path.basename(_src_path)
-    src_path = Path(_src_path) / role
-    if not src_path.exists():
-        print(f"Error: {src_path} does not exists.")
+        src_owner = os.path.basename(src_path)
+    _meta_path = src_path / "meta/main.yml"
```
richm, line 972: Are you using the existence of `meta/main.yml` to determine if the directory contains a role?  If so, then according to Ansible the bare minimum for a role is `tasks/main.yml` - ansible-lint also uses this to determine if the directory is a role or not.

nhosoi: Ok. Let me replace it with tasks. I thought it is common to check meta/main.yml looking at this line. :)
https://github.com/linux-system-roles/test-harness/blob/master/test/run-tests#L375

richm: that line is wrong too . . . https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/utils.py

nhosoi: Shall I open another PR for the line? or is it ok to include the fix in https://github.com/linux-system-roles/test-harness/pull/113?

richm: > Shall I open another PR for the line? or is it ok to include the fix in [linux-system-roles/test-harness#113](https://github.com/linux-system-roles/test-harness/pull/113)?

I think this PR

nhosoi: Ah, test-harness intentionally uses meta/main.yml to check meta info in checkmeta.py.
So, they are good with meta.

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #30, opened 2020-11-17
Collection release scripts, data, docs

richm:
see the README.md - release_collection.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ linux-system-roles repos.
   * [shellcheck](#shellcheck)
   * [sync-template.sh](#sync-templatesh)
   * [lsr_role2collection.py](#lsr_role2collectionpy)
+  * [release_collection.py](#release_collectionpy)
 <!--te-->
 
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -240,3 +241,129 @@ Destination collections path is /path/to/collections.
...
+build and publish.
+
+## collection_release.yml format
+```yaml
+ROLENAME:
+  ref: TAG_OR_HASH_OR_BRANCH
+  org: github-organization
+  repo: github-repo
+```
+Where `ROLENAME` is the name of the role as it will appear in the collection (under `NAMESPACE/COLLECTION_NAME/roles/`).  `ref` is the git tag (preferred), commit hash, or branch to use (in the format used as the argument to `git clone -b` or `git checkout`).  `org` is the github organization (default `linux-system-roles`).  The `repo` is the name of the repo under the github organization, and the default is the `ROLENAME`, so only use this if you need to specify a different name for the role in the collection.
```
nhosoi, line 293: > The `repo` is the name of the repo under the github organization, and the default is the `ROLENAME`

Just out of curiosity, can we have a repo which name is not identical to `ROLENAME`?

I'm afraid if we do, lsr_role2collection would not work in that case. It expects the directory name in COLLECTION_SRC_PATH (e.g., ~/linux-system-roles) is `ROLENAME`...

richm: Yes.  We are planning to release the ansible-sshd role with out collection.  The upstream for this is https://github.com/willshersystems/ansible-sshd.  It is included in the collection under the name `sshd` - https://galaxy.ansible.com/fedora/linux_system_roles
The script clones the repo under the name `sshd` using `git clone https://github.com/willshersystems/ansible-sshd /path/to/sshd` so that the `lsr_role2collection.py` script only sees the `sshd` name.

nhosoi: I see. This is a good example.
```
sshd:
  org: willshersystems
  repo: ansible-sshd
  ref: v0.12.0
```
To convert sshd to the collection format by `lsr_role2collection`, we have to clone `sshd` as
```
git clone https://github.com/willshersystems/ansible-sshd sshd
```

richm: The script does it here https://github.com/linux-system-roles/auto-maintenance/pull/30/files#diff-a7a7d93dac329fce31220c456b0b9cf11c19bfbe2fc0f9482e8c1366821760e7R56-R68 - the final argument `roledir` in this case ends with `/sshd`

nhosoi (approved):
lgtm

richm:
@pcahyna will this fit with the work you are doing with fedora/packit?

PR #31, opened 2020-11-18
Support include_vars in the test playbooks

nhosoi:
If a test playbook contains a relative path such as
```
  include_vars: relative_path/to/linux-system-roles.ROLENAME/file_or_dir
```
or
```
  include_vars:
    file|dir: relative_path/to/linux-system-roles.ROLENAME/file_or_dir
```
they need to be converted to the absolute path in the collection
format since the test playbook could be located at any place.
According to the include_vars doc, "If the path is relative and not
inside a role, it will be parsed relative to the playbook." To solve
the issue, the relative path is converted to the absolute path.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -35,6 +35,7 @@ import textwrap
 from pathlib import Path
 from ruamel.yaml import YAML
 from shutil import copytree, copy2, ignore_patterns, rmtree
+from six import string_types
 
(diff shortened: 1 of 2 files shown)


richm (approved):
lgtm

nhosoi:
Thanks for reviewing the PR, @richm. Merging...

PR #32, opened 2020-11-18
RHELPLAN-59755 - Collection - script - not include symlinks in the collection converted format

nhosoi:
lsr_role2collection.py - In copying the roles tests dir to the collections
tests, by setting symlinks=False, it follows the links and copies the real
files and directories. Plus, when calling copy_tree_with_replace, set
"linux-system-roles.ROLENAME" to "ignoreme", which avoids copying the dir.

Removing unnecessary test case from test_lsr_role2collection.py

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -591,7 +591,7 @@ def copy_tree_with_replace(
                 lsr_copytree(
                     src,
                     dest,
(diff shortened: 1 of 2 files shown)


richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merging.

PR #33, opened 2020-12-08
RHELPLAN-59755 - Collection - script - not include symlinks in the co…

nhosoi:
In the commit 6442eafcd69c91808ef6e11daf09d5c53d541f8d, symlinks in the
tests dir are converted to the real files and directories including
the module_utils, which is not preferable. With this commit, such sym-
links targeting the outside of the tests are removed. Only symlinks
which points inside the tests are untouched. For instance,
lrwxrwxrwx. <<snip>> network/tests/playbooks/files -> ../files/

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -625,7 +625,7 @@ def copy_tree_with_replace(
     transformer_args,
     isrole=True,
     ignoreme=None,
-    symlinks=False,
+    symlinks=True,
 ):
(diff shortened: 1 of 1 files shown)


richm (dismissed):
lgtm

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #35, opened 2020-12-11
RHELPLAN-59755 - Collection - script - not include symlinks in the collection converted format

nhosoi:
Fixing a copying symlink bug.
Function lsr_copytree should keep symlinks if option symlinks is True;
copy its contents if option symlinks is False. There was a bug to check
symlink in lsr_copytree. Path.is_dir returns true for symlink dir and
it's treated as a directory.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -34,7 +34,7 @@ import textwrap
 
 from pathlib import Path
 from ruamel.yaml import YAML
-from shutil import copytree, copy2, ignore_patterns, rmtree
+from shutil import copytree, copy2, copyfile, ignore_patterns, rmtree
 from six import string_types
 
 from ansible.errors import AnsibleParserError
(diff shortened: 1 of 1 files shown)


richm:
There is one place where we want to preserve symlinks.  In the `vars/` directory, it is quite common to have something like this:
```
vars/
  RedHat-8.yml
  CentOS-8.yml -> RedHat-8.yml
```
That is - the role needs to set specific variables for rhel8 or centos8 - but rather than have two separate files that are otherwise identical, the role will use RedHat-8.yml as the "real" file, and CentOS-8.yml is just a link to RedHat-8.yml

nhosoi:
> There is one place where we want to preserve symlinks. In the `vars/` directory, it is quite common to have something like this:
> 
> ```
> vars/
>   RedHat-8.yml
>   CentOS-8.yml -> RedHat-8.yml
> ```
> 
> That is - the role needs to set specific variables for rhel8 or centos8 - but rather than have two separate files that are otherwise identical, the role will use RedHat-8.yml as the "real" file, and CentOS-8.yml is just a link to RedHat-8.yml

Right. I have no plan to touch the symlinks under the `roles` dir. It is for the `tests` dir. And this change is rather for fixing a bug in the previous commit 9cf9aaeb49e72f5096e79a0f5866a2a178b9b18c...

richm (approved):
lgtm

nhosoi:
Thanks for your reviewing this pr, @richm.

PR #39, opened 2021-01-20
lsr_role2collections: Instead of hardcoded "linux-system-roles", use the src-owner value in FQRN.

nhosoi:
In the rpm packaging process, the "linux-system-roles" part in FQRN
could be replaced with the package name, which is equivalent to the
src-owner value.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -431,7 +431,7 @@ class LSRFileTransformer(LSRFileTransformerBase):
             To solve it, the relative path is converted to the absolute path.
             """
             _src_owner_match = "/" + self.src_owner + "."
-            _src_owner_pattern = r".*/{0}.(\w+)/([\w\d\./]+)".format(self.src_owner)
+            _src_owner_pattern = r".*/{0}[.](\w+)/([\w\d\./]+)".format(self.src_owner)
             if isinstance(task[module_name], dict):
                 _key = None
                 if (
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -15,7 +15,7 @@
 #                        [-h]
 # Or
 #
-# COLLECTION_SRC_PATH=/path/to/linux-system-roles \
+# COLLECTION_SRC_PATH=/path/to/{src_owner} \
 # COLLECTION_DEST_PATH=/path/to/collections \
 # COLLECTION_NAMESPACE=mynamespace \
 # COLLECTION_NAME=myname \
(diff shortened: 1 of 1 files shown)


richm:
This assumes that if you have a role with source like `https://github.com/orgname/reponame` and it uses `include_vars`, it uses `include_vars` like this:
```yaml
include_vars:
  file: some/path/to/orgname/ROLENAME/file.yml
```
and the script will convert this to
```yaml
include_vars:
  file: /usr/share/ansible/collections/ansible_collections/fedora/linux_system_roles/ROLENAME/file.yml
```
?

nhosoi:
> This assumes that if you have a role with source like `https://github.com/orgname/reponame` and it uses `include_vars`, it uses `include_vars` like this:
> 
> ```yaml
> include_vars:
>   file: some/path/to/orgname/ROLENAME/file.yml
> ```
> 
> and the script will convert this to
> 
> ```yaml
> include_vars:
>   file: /usr/share/ansible/collections/ansible_collections/fedora/linux_system_roles/ROLENAME/file.yml
> ```
> 
> ?

Sorry, could you give me some more details? Are you suggesting to add the feature???

richm:
> Sorry, could you give me some more details? Are you suggesting to add the feature???

I'm trying to figure out what this PR is doing.

nhosoi:
The issue I'm trying to solve is including role in the test playbook like this
```
  roles:
    - linux-system-roles.ROLENAME
```
is replaced to
```
  roles:
    - rhel-system-roles.ROLENAME
```
using `sed` in the build spec file. The lsr_role2collection.py was checking the hardcoded `linux-system-roles` to determine whether to replace the value with FQCN or not, which does not work once `linux-system-roles` is replaced with something else, e.g., `rhel-system-roles`.

To make it work, we have to check FQRN with the --src-owner value instead of hardcoded `linux-system-roles`...

When the --src-owner option was added, it was meant to support a case the linux-system-roles/ROLENAME was checked out in a random dir other than linux-system-roles as being done in the run-tests script. I think we need to extend it to cover this non-linux-system-roles case...

pcahyna:
Is there any problem with the change? Removal of hardcoded values looks like a benefit to me, but maybe I am underestimating the effort needed to make the change work properly.

richm:
> Is there any problem with the change? Removal of hardcoded values looks like a benefit to me, but maybe I am underestimating the effort needed to make the change work properly.

Yes, it is a good idea to eliminate hardcoded values, but it wasn't necessary to solve the particular problem being faced.  So yes, let's reopen this PR.

nhosoi:
@pcahyna, @richm, could you please review this pr one more time?
This pr blocks the collection packaging.

lsr_role2collection.py
```diff
@@ -422,31 +422,31 @@ def task_cb(self, task):
...
             or
             include_vars:
-              file|dir: path/to/linux-system-roles.ROLENAME/file_or_dir
+              file|dir: path/to/{src_owner}.ROLENAME/file_or_dir
             Note: If the path is relative and not inside a role,
             it will be parsed relative to the playbook.
             To solve it, the relative path is converted to the absolute path.
             """
+            _src_owner_match = "/" + self.src_owner + "."
+            _src_owner_pattern = r".*/{0}.(\w+)/([\w\d\./]+)".format(self.src_owner)
```
richm, line 434: I think this should be `r".*/{0}[.](\w+)/([\w\d\./]+)".format(self.src_owner)` - that is, we want the `.` between `{0}` and `(\w+)` to match the literal dot `.` character - otherwise, it will match any character.

nhosoi: Thank you, @richm. Fixed.

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #45, opened 2021-02-03
Replacing "a collection of roles" with "a set of roles" in lsr_role2collection/collection_readme.md

nhosoi:
Per discussion in the meeting, we'd better avoid using the word `collection` to reduce the confusion with `ansible collections`.


richm (approved):

nhosoi:
Thank you, @richm !

PR #49, opened 2021-02-12
Commenting out empty dependencies in galaxy.yml

nhosoi:
It is for fixing this failure in the packaging.
```
Traceback (most recent call last):
  File "/usr/lib/rpm/ansible-generator", line 46, in <module>
    main()
  File "/usr/lib/rpm/ansible-generator", line 31, in main
    for dep, req in info.get("dependencies", {}).items():
AttributeError: 'NoneType' object has no attribute 'items'
```

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -20,7 +20,7 @@ license:
     - MIT
     - BSD-3-Clause
 
-dependencies:
+# dependencies: []
 
 tags:
   - "system"


richm (approved):
lgtm

nhosoi:
Thank you, @richm!

galaxy.yml
```diff
@@ -20,7 +20,7 @@ license:
     - MIT
     - BSD-3-Clause
 
-dependencies:
```
pcahyna, line 23: Just a note, are both changes needed? I suspect only one of them would be enough, i.e. either comment out the line or replacing the empty value by an empty list.

nhosoi: > Just a note, are both changes needed? I suspect only one of them would be enough, i.e. either comment out the line or replacing the empty value by an empty list.

I was hoping to give at least an idea of the type of the values. But it turned out my attempt was not valid (it's not a simple list.) I should have done something like this? It'd be needed for metrics in the next version, I guess.
```
# dependencies:
#     "other_namespace.collection1": ">=1.0.0"
#     "other_namespace.collection2": ">=2.0.0,<3.0.0"
```

pcahyna: Ah yes, right, it is a dict, not a list. I suggest to do exactly what ansible-collection does when creating a new collection:
```sh
ansible-galaxy collection init my_namespace.my_collection
```
galaxy.yml contains:
```yaml
# Collections that this collection requires to be installed for it to be usable. The key of the dict is the
# collection label 'namespace.name'. The value is a version range
# L(specifiers,https://python-semanticversion.readthedocs.io/en/latest/#requirement-specification). Multiple version
# range specifiers can be set and are separated by ','
dependencies: {}
```

nhosoi: > Ah yes, right, it is a dict, not a list. I suggest to do exactly what ansible-collection does when creating a new collection:
> 
> ```shell
> ansible-galaxy collection init my_namespace.my_collection
> ```
> 
> galaxy.yml contains:
> 
> ```yaml
> # Collections that this collection requires to be installed for it to be usable. The key of the dict is the
> # collection label 'namespace.name'. The value is a version range
> # L(specifiers,https://python-semanticversion.readthedocs.io/en/latest/#requirement-specification). Multiple version
> # range specifiers can be set and are separated by ','
> dependencies: {}
> ```

Thanks, @pcahyna. I've opened a [pr/50](https://github.com/linux-system-roles/auto-maintenance/pull/50). Having the right type (even if it is empty), it does not fail as reported in https://github.com/linux-system-roles/auto-maintenance/pull/49#issue-572239491.

PR #50, opened 2021-02-15
Fixing the dependencies value type in galaxy.yml

nhosoi:
Note:
Collections that this collection requires to be installed for it to be
usable. The key of the dict is the collection label 'namespace.name'.
The value is a version range
L(specifiers,https://python-semanticversion.readthedocs.io/en/latest/#requirement-specification).
Multiple version range specifiers can be set and are separated by ','

dependencies: {}

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -20,7 +20,7 @@ license:
     - MIT
     - BSD-3-Clause
 
-# dependencies: []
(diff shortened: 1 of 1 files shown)


pcahyna (approved):

nhosoi:
Thank you, @pcahyna!

PR #55, opened 2021-03-16
changes for collection release 0.0.3

richm:
Sync with the latest downstream usage of lsr_role2collection.
The release_collection.py script will construct the
`tests/sanity/ignore-2.9.txt` based on the ignore files from
the roles and the `lsr_role2collection/extra-ignore-2.9.txt`.
Run `galaxy-importer` against the collection package file for
additional checking if galaxy import will succeed.


nhosoi:
LGTM

I have one question about "lsr_role2collection/ignore-2.9.txt". 

The downstream does not use the file (at least in the current release). It only depends on .sanity-ansible-ignore-2.9.txt in each role.

Does the upstream use 3 of them if any? ignore-2.9.txt, extra- ignore-2.9.txt and .sanity-ansible-ignore-2.9.txt files? If so, how do we choose the 2 ignore files ignore-2.9 and extra-ignore-2.9 in auto-maintenance? ignore-2.9 is more basic and not touched often, but you could add any skip to extra-ignore? I'm wondering if we could have just one ignore file to simplify the steps...

richm:
> LGTM
> 
> I have one question about "lsr_role2collection/ignore-2.9.txt".
> 
> The downstream does not use the file (at least in the current release). It only depends on .sanity-ansible-ignore-2.9.txt in each role.
> 
> Does the upstream use 3 of them if any? ignore-2.9.txt, extra- ignore-2.9.txt and .sanity-ansible-ignore-2.9.txt files?

Upstream uses extra-ignore-2.9.txt and .sanity-ansible-ignore-2.9.txt files.  It does not use ignore-2.9.txt any more.

> If so, how do we choose the 2 ignore files ignore-2.9 and extra-ignore-2.9 in auto-maintenance? ignore-2.9 is more basic and not touched often, but you could add any skip to extra-ignore? I'm wondering if we could have just one ignore file to simplify the steps...

I suppose we can get rid of ignore-2.9.txt.

nhosoi (approved):
lgtm

PR #56, opened 2021-03-16
Bug 1935451 - Updates for the Automation Hub release

nhosoi:
https://bugzilla.redhat.com/show_bug.cgi?id=1935451#c10
- Needs to remove .dev-tools/ in the collection packaging.
- contributing.md that refers the directory is also removed from the
  collection.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -365,6 +365,8 @@ DO_NOT_COPY = (
     "Vagrantfile",
     "README.html",
     "CHANGELOG",
+    "contributing.md",
+    ".dev-tools",
 )
 
 ALL_DIRS = ROLE_DIRS + PLUGINS + TESTS + DOCS + DO_NOT_COPY


nhosoi:
Closing since this is not what we want...

pcahyna:
I am surprised. I thought that the RPM build process simply ignores all files starting with dots, but now I see plenty of special cases instead. Wasn't there a proposed change that removed all the dotfiles? I can't find it in the spec file now, but I would swear that I saw it.

pcahyna:
> I am surprised. I thought that the RPM build process simply ignores all files starting with dots, but now I see plenty of special cases instead. Wasn't there a proposed change that removed all the dotfiles? I can't find it in the spec file now, but I would swear that I saw it.

I found it. The cleanup of dotfiles and other unnecessary/development files happens here:
https://src.fedoraproject.org/rpms/linux-system-roles/blob/rawhide/f/linux-system-roles.spec#_516

The problem is, it happens in the installed role, so it is not propagated to the collection and the conversion tool has to reimplement it.

richm:
> > I am surprised. I thought that the RPM build process simply ignores all files starting with dots, but now I see plenty of special cases instead. Wasn't there a proposed change that removed all the dotfiles? I can't find it in the spec file now, but I would swear that I saw it.
> 
> I found it. The cleanup of dotfiles and other unnecessary/development files happens here:
> https://src.fedoraproject.org/rpms/linux-system-roles/blob/rawhide/f/linux-system-roles.spec#_516
> 
> The problem is, it happens in the installed role, so it is not propagated to the collection and the conversion tool has to reimplement it.

The conversion tool may be run outside of a spec file e.g. for galaxy we may want to run the role2collection conversion directly against a git clone of an upstream role repo.

richm:
I also don't like having all of these special cases - perhaps we should use an include list and just say "if the file/directory is not in the set of role files/directories like 'README.md', 'tasks', 'vars', etc. then don't copy/convert it"?

pcahyna:
@richm I see. What if we add a Makefile that would allow to do "make install" that would copy all the desired files to a destination? This "make install" could be then used both in the spec file and outside the spec file and both for preparing files for a collection and for installing the legacy roles?

pcahyna:
Note that using "make install" in spec has been my secret dream for some time already, regardless of this particular issue. The way it is done now in the spec is quite clumsy.

nhosoi:
@richm, @pcahyna, So, for the short term goal, what shall we do?
Is it ok to close this pr and remove dot files/dirs in the spec file for now?
If time allows we could introduce a makefile to define the file set to release/package? (maybe, in the next release?)

richm:
> @richm, @pcahyna, So, for the short term goal, what shall we do?
> Is it ok to close this pr and remove dot files/dirs in the spec file for now?

Yes

> If time allows we could introduce a makefile to define the file set to release/package? (maybe, in the next release?)

Yes, in another PR

nhosoi:
Closing this pr per discussion.

PR #57, opened 2021-03-16
md2html.sh - adding an option -l

nhosoi:
Usage: md2html.sh [-h] [[-l] md_file0 md_file1 ...]
If "-l" is set and if an option md file contains a link to
other md file in it, convert the link to html.


richm (approved):
lgtm

PR #60, opened 2021-04-05
md2html.sh - Make sure there is no CRLF

nhosoi:


richm (approved):
lgtm

pcahyna:
Why is the .html file part of the collection artifact for Galaxy/Automation Hub? I thought that Automation Hub renders the original .md files and does not care about .html. Would it be possible to build the .html files only after building the collection artifact?
(One effect would be that collection installed from AH or Galaxy would not have the .html files. Is it usual to have such files or not?)

nhosoi:
> Why is the .html file part of the collection artifact for Galaxy/Automation Hub? I thought that Automation Hub renders the original .md files and does not care about .html. Would it be possible to build the .html files only after building the collection artifact?
> (One effect would be that collection installed from AH or Galaxy would not have the .html files. Is it usual to have such files or not?)

This is for fixing these errors in the Automation Hub import.
```
ERROR: README.html:0:0: use "\n" for line endings instead of "\r\n" 
ERROR: roles/certificate/README.html:0:0: use "\n" for line endings instead of "\r\n" 
ERROR: roles/crypto_policies/README.html:0:0: use "\n" for line endings instead of "\r\n" 
ERROR: roles/ha_cluster/README.html:0:0: use "\n" for line endings instead of "\r\n" 
ERROR: roles/kdump/README.html:0:0: use "\n" for line endings instead of "\r\n" 
8< ------------------------------------------------------------------------ >8
```

pcahyna:
> > Why is the .html file part of the collection artifact for Galaxy/Automation Hub? I thought that Automation Hub renders the original .md files and does not care about .html. Would it be possible to build the .html files only after building the collection artifact?
> > (One effect would be that collection installed from AH or Galaxy would not have the .html files. Is it usual to have such files or not?)
> 
> This is for fixing these errors in the Automation Hub import.
> 
> ```
> ERROR: README.html:0:0: use "\n" for line endings instead of "\r\n" 
> ERROR: roles/certificate/README.html:0:0: use "\n" for line endings instead of "\r\n" 
> ERROR: roles/crypto_policies/README.html:0:0: use "\n" for line endings instead of "\r\n" 
> ERROR: roles/ha_cluster/README.html:0:0: use "\n" for line endings instead of "\r\n" 
> ERROR: roles/kdump/README.html:0:0: use "\n" for line endings instead of "\r\n" 
> 8< ------------------------------------------------------------------------ >8
> ```

I know and I am pointing out that by leaving out the .html files from the collection, the errors would be solved. My question was whether that would be a good solution.

nhosoi:
> I know and I am pointing out that by leaving out the .html files from the collection, the errors would be solved. My question was whether that would be a good solution.

??? Aren't we supposed to convert the README.md files to the html format and package them? I think it was pointed out by @xjezda00 in his verification effort and we fixed in this [bz](https://bugzilla.redhat.com/show_bug.cgi?id=1935451#c10).
`6) The collection READMEs are not being converted to html nor copied / linked to /usr/share/doc/rhel-system-roles. Nor tagged as docfiles in the RPM metadata.`

richm:
> > I know and I am pointing out that by leaving out the .html files from the collection, the errors would be solved. My question was whether that would be a good solution.
> 
> ??? Aren't we supposed to convert the README.md files to the html format and package them? I think it was pointed out by @xjezda00 in his verification effort and we fixed in this [bz](https://bugzilla.redhat.com/show_bug.cgi?id=1935451#c10).
> `6) The collection READMEs are not being converted to html nor copied / linked to /usr/share/doc/rhel-system-roles. Nor tagged as docfiles in the RPM metadata.`

Yes.  The html format files are supposed to be in the *rpm package*.  The question is - do we need them in the *collection* for Automation Hub?

nhosoi:
> Yes. The html format files are supposed to be in the _rpm package_. The question is - do we need them in the _collection_ for Automation Hub?

Ah, ok. Let me try dropping them and importing to Automation Hub. Thanks, @pcahyna and @richm.

PR #61, opened 2021-04-07
For .md to .adoc conversion, substitute pandoc with kramdoc

spetrosi:
Bug 1943532 - Remove dependency on pandoc
https://bugzilla.redhat.com/show_bug.cgi?id=1943532

### DIFF
--- a/md2html.sh
+++ b/md2html.sh
@@ -29,7 +29,7 @@ for file in "$@"; do
         # With kramdown, convert directly to HTML
         if [ $md2html_tool == kramdown ]; then
           # Set locale to UTF-8 because by default it is set to US-ASCII
-          LANG=en_US.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
+          LC_ALL=en_US.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
           --output html "${file}" > "${file%.md}.html"
         # With pandoc, convert to adoc, then to HTML
         elif [ $md2html_tool == pandoc ]; then
--- a/md2html.sh
+++ b/md2html.sh
@@ -1,49 +1,52 @@
 #!/bin/bash
 
 set -euxo pipefail
+
 convert_link=0
+
 while getopts "hl" opt; do
-	case "$opt" in
-		h)
-			echo "Usage: $0 [-h] [[-l] md_file0 md_file1 ...]"
-			echo "options:"
-			echo "-l  convert links to other md files to html"
-			exit ;;
-		l)
-			convert_link=1
-			shift ;;
-		*)
-			;;
-	esac
+  case "$opt" in
+    h)
+      echo "Usage: $0 [-h] [[-l] md_file0 md_file1 ...]"
+      echo "options:"
+      echo "-l  convert links to other md files to html"
+      exit ;;
+    l)
+      convert_link=1
+      shift ;;
+    *)
+      ;;
+  esac
 done
 
 for file in "$@"; do
-	md2html_tool=""
-	# RHEL 9 in brew cannot use pandoc, hence trying kramown first
-	if type -p kramdown >/dev/null && kramdown -i GFM < /dev/null > /dev/null; then
-	    md2html_tool=kramdown
-	elif type -p pandoc >/dev/null; then
-	    md2html_tool=pandoc
-	fi
-	# With kramdown, convert directly to HTML
-	if [ $md2html_tool == kramdown ]; then
-	    # Set locale to UTF-8 because by default it is set to US-ASCII
-	    LC_ALL=C.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
-	    --output html "${file}" > "${file%.md}.html"
-	# With pandoc, convert to adoc, then to HTML
-	elif [ $md2html_tool == pandoc ]; then
-	    $md2html_tool -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
-	    touch -r "${file}" "${file%.md}.tmp.adoc"
-	    TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
-	    tr -d '\r' < "${file%.md}.html" > "${file%.md}.tmp.adoc"
-	    mv "${file%.md}.tmp.adoc" "${file%.md}.html"
-	else
-	    echo "Cannot find a tool to convert md to adoc"
-	    echo "You must install rubygem-kramdown-parser-gfm or pandoc"
-	    exit 1
-	fi
-	# Convert links
-	if [ "$convert_link" -ne 0 ]; then
-	    sed -i -e "s/\.md\>/\.html/g" "${file%.md}.html"
-	fi
+  md2html_tool=""
+  # RHEL 9 in brew cannot use pandoc, hence trying kramown first
+  if type -p kramdown >/dev/null && kramdown -i GFM < /dev/null > /dev/null; then
+    md2html_tool=kramdown
+  elif type -p pandoc >/dev/null; then
+    md2html_tool=pandoc
+  fi
+  # With kramdown, convert directly to HTML
+  if [ $md2html_tool == kramdown ]; then
+    # Set locale to UTF-8 because by default it is set to US-ASCII
+    LC_ALL=C.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
+    --output html "${file}" > "${file%.md}.html"
+  # With pandoc, convert to adoc, then to HTML
+  elif [ $md2html_tool == pandoc ]; then
+    $md2html_tool -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+    touch -r "${file}" "${file%.md}.tmp.adoc"
+    TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
+    tr -d '\r' < "${file%.md}.html" > "${file%.md}.tmp.adoc"
+    mv "${file%.md}.tmp.adoc" "${file%.md}.html"
+  else
+    echo "Cannot find a tool to convert md to adoc"
+    echo "You must install rubygem-kramdown-parser-gfm or pandoc"
+    exit 1
+  fi
+
+  # Convert links
+  if [ "$convert_link" -ne 0 ]; then
+    sed -i -e "s/\.md\>/\.html/g" "${file%.md}.html"
+  fi
 done
--- a/md2html.sh
+++ b/md2html.sh
@@ -20,7 +20,7 @@ done
 for file in "$@"; do
         md2html_tool=""
         # RHEL 9 in brew cannot use pandoc, hence trying kramown first
-        if type -p kramdown >/dev/null && LANG=en_US.UTF-8 kramdown -i GFM "${file}">/dev/null; then
+        if type -p kramdown >/dev/null && kramdown -i GFM < /dev/null > /dev/null; then
           md2html_tool=kramdown
         elif type -p pandoc >/dev/null; then
           md2html_tool=pandoc
--- a/md2html.sh
+++ b/md2html.sh
@@ -6,6 +6,8 @@ while getopts "hl" opt; do
 	case "$opt" in
 		h)
 			echo "Usage: $0 [-h] [[-l] md_file0 md_file1 ...]"
+                        echo "options:"
+                        echo "-l  convert links to other md files to html"
 			exit ;;
 		l)
 			convert_link=1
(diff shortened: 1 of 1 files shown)


pcahyna:
I think you should try both - there probably are versions of RHEL that don't have kramdoc and have pandoc.

spetrosi:
Together with this PR I need to also add the req for the kramdoc rubygem to the spec file

spetrosi:
> I think you should try both - there probably are versions of RHEL that don't have kramdoc and have pandoc.

We are supposed to remove this dependency on pandoc, that's the thing. What system build RPMs in brewweb? I though that is always uses some RHEL to build packages for all OSes.

pcahyna:
> > I think you should try both - there probably are versions of RHEL that don't have kramdoc and have pandoc.
> 
> We are supposed to remove this dependency on pandoc, that's the thing. 

That's not correct. We are supposed to remove this dependency in RHEL 9, not others.

> What system build RPMs in brewweb? I though that is always uses some RHEL to build packages for all OSes.

The RPMs are built against the buildroot of the RHEL version that the RPMs are intended for. RHEL 7 packages against RHEL 7 buildroot, RHEL 8 packages against RHEL 8 buildroot, etc. Think about it, it would not work otherwise.

spetrosi:
> The RPMs are built against the buildroot of the RHEL version that the RPMs are intended for. RHEL 7 packages against RHEL 7 buildroot, RHEL 8 packages against RHEL 8 buildroot, etc. Think about it, it would not work otherwise.
 
Thank you, I made kramdoc run on RHEL 9 only, and pandoc on other RHEL. How to find out if the `rubygem-kramdown` package is available on the buildroot RHELs? I can find neither `pandoc` nor `rubygem-kramdown` on RHELs now in rhel-baseos and appstream repos.

md2html.sh
```diff
@@ -16,7 +16,11 @@ while getopts "hl" opt; do
 done
 
 for file in "$@"; do
-	pandoc -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+        if grep -qi 'Red Hat Enterprise Linux release 9' /etc/redhat-release; then
```
pcahyna, line 19: Please, don't do this.
First, checking for a specific release is very fragile. You don't have any guarantee that the OS the script will run on will be named exactly like that and contain/not contain the package that you need. It is much better to check for the actual thing that you need. i.e. pandoc or kramdoc. One example (for deciding between asciidoc and asciidoctor) can be found here: https://github.com/rear/rear/blob/6a3d0b4d5e73c69a62ce0bd209b2b38ffb462569/doc/Makefile#L10
Second, even in cases when you need to determine the OS version and release, there are much more structured approaches to that than searching for a string in a file. See os-release(5).

spetrosi: Thanks a lot, I was feeling that reading strings in the file is bad. Thanks for the explanation, I've tried another thing now.

md2html.sh
```diff
@@ -16,7 +16,13 @@ while getopts "hl" opt; do
 done
 
 for file in "$@"; do
-	pandoc -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+        # RHEL 9 in brew cannot use pandoc, hence trying kramdoc first
+        if kramdoc --format=GFM --output="${file%.md}.tmp.adoc" "${file}"; then
```
pcahyna, line 20: This way a failure of kramdoc will be treated in exactly the same way as failure due to kramdoc not found. If kramdoc fails because of some problem with the file, pandoc will be attempted next, and usually will also fail, because usually only one of pandoc and kramdoc will be installed. This will lead to a very confusing error because you will think that the problem is that pandoc is not found, while the real problem would be the kramdoc failure above.
I suggest to first determine whether to use pandoc or kramdoc using a method similar to the one ReaR is using, and stick with the choice.

spetrosi: Completely agree now, done that

spetrosi:
We’ve found out that the  kramdoc command is not included with the rubygem-kramdown RPM in Fedora, spetrosi will try using plain kramdown instead

md2html.sh
```diff
@@ -16,7 +16,22 @@ while getopts "hl" opt; do
 done
 
 for file in "$@"; do
-	pandoc -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+        md2adoc_tool=""
+        # RHEL 9 in brew cannot use pandoc, hence trying kramdoc first
+        if type -p kramdown >/dev/null; then
```
pcahyna, line 21: Here you try `kramdown` - but below yiu actually use `kramdoc`, not `kramdown`.

md2html.sh
```diff
@@ -16,7 +16,22 @@ while getopts "hl" opt; do
...
+          md2adoc_tool=kramdown
```
pcahyna, line 22: Is this variable ever used?

md2html.sh
```diff
@@ -16,7 +16,22 @@ while getopts "hl" opt; do
...
+        elif type -p pandoc >/dev/null; then
+          md2adoc_tool=pandoc
+        fi
+        if [ $md2adoc_tool == kramdown ]; then
+          kramdoc --format=GFM --output="${file%.md}.tmp.adoc" "${file}"
```
pcahyna, line 27: No Fedora package contains `kramdoc`, so this will be problematic to use on Fedora/RHEL.

md2html.sh
```diff
@@ -16,12 +18,36 @@ while getopts "hl" opt; do
...
-	fi
-	touch -r "${file}" "${file%.md}.tmp.adoc"
-	TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
-	tr -d '\r' < "${file%.md}.html" > "${file%.md}.tmp.adoc"
-	mv "${file%.md}.tmp.adoc" "${file%.md}.html"
+        md2html_tool=""
+        # RHEL 9 in brew cannot use pandoc, hence trying kramown first
+        if rpm -q rubygem-kramdown-parser-gfm > /dev/null; then
+          md2html_tool=kramdown
+        elif rpm -q pandoc >/dev/null; then
```
pcahyna, line 25: I think it is better to check for the availability of the actual commands, rather than for the presence of the RPMs that contain them, because you ultimately care about the commands, not the RPMs. Repeating an example: https://github.com/rear/rear/blob/6a3d0b4d5e73c69a62ce0bd209b2b38ffb462569/doc/Makefile#L10

spetrosi: I remember about checking with `type`, however, the issue here is that for the kramdown, the script needs the kramdown-parser-gfm RPM installed, not just kramdown. I can check it by verifying the files that the `rubygem-kramdown-parser-gfm` package or the `kramdown-parser-gfm` gem installs, however checking files existance does not seem as reliable as checking for the RPM. Does that sound correct?
I then used the same check for the rpm for pandoc too intuitively for consistency, however, I can use `type` for pandoc as before, will that help?

pcahyna: Hmm, can you perform a dummy invocation of `kramdown` that will verify if the gfm parser is installed?

pcahyna: apparently this won't work, because if the gfm parser is not installed, `kramdown -x parser-gfm` will just print the message `Couldn't load extension parser-gfm, ignoring` and continue as if nothing wrong has happened. `gem info kramdown-parser-gfm` also does not return any error if the gem is not available. So keep the code as it is now because I don't see any better option.

md2html.sh
```diff
@@ -16,12 +18,36 @@ while getopts "hl" opt; do
...
+          md2html_tool=pandoc
+        fi
+
+        # With kramdown, convert directly to HTML
+        if [ $md2html_tool == kramdown ]; then
+          # Run kramdown with `ruby` to specify the encoding as ruby uses US-ASCII by default
+          ruby --encoding=UTF-8 -S $md2html_tool --extension parser-gfm \
```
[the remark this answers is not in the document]
spetrosi: Yes, e.g. run `kramdown -i GFM README.md`, this either passes with `0` rc, or fails with `Error: kramdown has no parser to handle the specified input format: GFM` and `1` rc.
I will use this instead of the `rpm -ql` check. Thank you

md2html.sh
spetrosi: Yes, e.g. run kramdown -i GFM README.md, this either passes with 0 rc, or fails with Error: kramdown has no parser to handle the specified input format: GFM and 1 rc.
I will use this instead of the rpm -ql check. Thank you

md2html.sh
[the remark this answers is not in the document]
spetrosi: As per our IRC discussion, I will use `LC_ALL=en_US.UTF-8 ruby` as you suggested. Thank you

md2html.sh
pcahyna: good, I missed the difference between `-i` and `-x`. You can use `kramdown -i GFM < /dev/null`, no need to use a real file.

md2html.sh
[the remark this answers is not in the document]
spetrosi: As per our discussion, I used `LC_ALL=C.UTF-8` with the kramdown command, this works on Fedora just fine:
https://koji.fedoraproject.org/koji/taskinfo?taskID=67821072
This will also work on RHEL 9. This won't work on RHEL 7, but RHEL 7 will use pandoc and asciidoctor, not kramdown.

richm:
The formatting and indentation are not correct.

In the old code the convert_link step was between the pandoc and the asciidoc steps - in the new code it is not.  Is this correct?

spetrosi:
> The formatting and indentation are not correct.

GitHub displays indentation incorrectly, not sure if it is wrong in this PR with shell scripts in common. Please review changes locally to check indendation.
> 
> In the old code the convert_link step was between the pandoc and the asciidoc steps - in the new code it is not. Is this correct?

I checked this by running the script for sap-hana-preconfigure readmes, the main README.md includes a link to the README.DEV.md. The following command converts the readmes correctly, i.e. the link to README.DEV.md works fine, both using kramdown and pandoc.
`sh md2html.sh -l ../sap-hana-preconfigure/README.md ../sap-hana-preconfigure/README.DEV.md`

pcahyna:
> > The formatting and indentation are not correct.
> 
> GitHub displays indentation incorrectly, not sure if it is wrong in this PR with shell scripts in common. Please review changes locally to check indendation.

Well I don't know what you mean by "with shell scripts in common", but it is not usual that GitHub web diffs show indentation incorrectly, so one should not be forced to review changes locally. Moreover, when reviewing locally, `git log -u` shows the same problem. So, it is better to ask what the problem is instead of sweeping it under the rug (in this case, the problem lies in inconsistent usage of tabs vs. spaces).

spetrosi:
> > > The formatting and indentation are not correct.
> > 
> > 
> > GitHub displays indentation incorrectly, not sure if it is wrong in this PR with shell scripts in common. Please review changes locally to check indendation.
> 
> Well I don't know what you mean by "with shell scripts in common", but it is not usual that GitHub web diffs show indentation incorrectly, so one should not be forced to review changes locally. Moreover, when reviewing locally, `git log -u` shows the same problem. So, it is better to ask what the problem is instead of sweeping it under the rug (in this case, the problem lies in inconsistent usage of tabs vs. spaces).

I did not have a chance to face such an issue before and had no idea of Git processing tabs and spaces differently. I think I've fixed this now. Thank you!

pcahyna:
yes, it looks good now. Before you merge, I think that it would be better to squash the commit that fixes the problem with the commit that introduced it in order to simplify the history.

By the way, maybe it is a good time to change tabs to spaces in this file, because as you have just seen, tabs are annoying (sorry for having introduced them, I suspect the file comes from me). But I will leave this style decision to @richm.

> had no idea of Git processing tabs and spaces differently

This may be a bit a misconception - git does nothing special here, the difference is inherent in the distinction between tabs and spaces (so any tool that displays text will show them differently, even if it just outputs the text on screen, because the terminal handles it).

richm:
> By the way, maybe it is a good time to change tabs to spaces in this file, because as you have just seen, tabs are annoying (sorry for having introduced them, I suspect the file comes from me). But I will leave this style decision to @richm.

yes, please change the tabs to spaces

> (so any tool that displays text will show them differently, even if it just outputs the text on screen, because the terminal handles it)

yes, this is the main reason why I prefer spaces, because tabs are rendered differently in each context, which is confusing

spetrosi:
I changed tabs to spaces in the last commit.

md2html.sh
```diff
@@ -1,27 +1,52 @@
...
-	mv "${file%.md}.tmp.adoc" "${file%.md}.html"
+  md2html_tool=""
+  # RHEL 9 in brew cannot use pandoc, hence trying kramown first
+  if type -p kramdown >/dev/null && kramdown -i GFM < /dev/null > /dev/null; then
+    md2html_tool=kramdown
+  elif type -p pandoc >/dev/null; then
+    md2html_tool=pandoc
+  fi
+  # With kramdown, convert directly to HTML
+  if [ $md2html_tool == kramdown ]; then
```
richm, line 31: use double quotes around `$md2html_tool` since its value could be `""` and then the `if` will evaulate to this:
```
if [ == kramdown ]
```
which will throw a strange error

md2html.sh
```diff
@@ -1,27 +1,52 @@
...
+    # Set locale to UTF-8 because by default it is set to US-ASCII
+    LC_ALL=C.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
+    --output html "${file}" > "${file%.md}.html"
+  # With pandoc, convert to adoc, then to HTML
+  elif [ $md2html_tool == pandoc ]; then
```
richm, line 36: see above

richm:
couple of minor nits, otherwise, lgtm

md2html.sh
spetrosi: Thanks, resolved this

md2html.sh
spetrosi: Done.

richm (approved):
lgtm

PR #63, opened 2021-04-16
Adding check_rpmspec_collection.sh

nhosoi:
Description
- Verify rpm package count with various combination of ansible and
              collection_artifact options.
- Check there is no README.html files in /usr/share/andible/collections.

Usage: ./check_rpmspec_collection.sh [ -h | --help ] | [ fedpkg | rhpkg [ branch_name ] ]

### DIFF
--- a/check_rpmspec_collection.sh
+++ b/check_rpmspec_collection.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+if [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+    echo "Description - Verify rpm package count with various combination of ansible and collection_artifact options."
+    echo "            - Check there is no README.html files in /usr/share/andible/collections."
+    echo "Usage: $0 [ -h | --help ] | [ fedpkg | rhpkg [ branch_name ] ]"
+    exit 0
+fi
+
+if [ "$( find . -maxdepth 1 -name '*.spec' | wc -l )" -ne 1 ]; then
+    echo There must exactly 1 spec file in "$( pwd )".
+    exit 1
+fi
+
+set -euo pipefail
+
+pkgcmd=${1:-"fedpkg"}
+branch=${2:-"rawhide"}
+run() {
+    local option=$1
+    local ext=$2
+    local expect=$3
+    rm -rf noarch
+    output=/tmp/$pkgcmd.$ext
+    cmdline="$pkgcmd --release $branch local $option"
+    echo "$cmdline"
+    $cmdline > "$output" 2>&1
+    count=$( find noarch -name "*.rpm" | wc -l )
+    if [ "$count" -eq "$expect" ]; then
+      echo OK - result "$count" is "$expect"
+    else
+      echo FAILED - result "$count" is not "$expect"
+    fi
+
+    filename=$( "$pkgcmd" --release "$branch" verrel )
+    if rpm -qlp noarch/"${filename}"*.rpm | grep -F '.html' | grep collection | grep -v '/usr/share/doc/'; then
+      echo FAILED - '.html' files are in noarch/"${filename}"*.rpm other than doc
+    else
+      echo OK - '.html' files are only in doc in noarch/"${filename}"*.rpm
+    fi
+}
+
+run "--without ansible" "without_ansible" 1
+run "--without ansible --with collection_artifact" "without_ansible-with_collection_artifact" 1
+run "--without ansible --without collection_artifact" "without_ansible-without_collection_artifact" 1
+run "--without collection_artifact" "without_collection_artifact" 1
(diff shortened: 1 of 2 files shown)


check_rpmspec_collection.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
```
richm, line 2: please use `set -euo pipefail`

check_rpmspec_collection.sh
```diff
@@ -0,0 +1,50 @@
...
+if [ "$1" = "-h" -o "$1" = "--help" ]; then
+    echo "Description - Verify rpm package count with various combination of ansible and collection_artifact options."
+    echo "            - Check there is no README.html files in /usr/share/andible/collections."
+    echo "Usage: $0 [ -h | --help ] | [ fedpkg | rhpkg [ branch_name ] ]"
+    exit 0
+fi
+
+if [ ! -f *.spec ]; then
```
richm, line 10: if there is more than one spec file you will get an error like this:
```
bash: [: another.spec: binary operator expected
```

richm: Maybe something like this:
```bash
if ls -1 *.spec | wc -l | grep -q '^1$'; then
  echo good
else
  echo there must exactly 1 spec file in $(pwd)
  exit 1
fi
```
if there are no spec files the output looks like this:
```
ls: cannot access '*.spec': No such file or directory
there must exactly 1 spec file in /home/rmeggins/fedpkg/linux-system-roles
```

check_rpmspec_collection.sh
```diff
@@ -0,0 +1,52 @@
...
+
+set -euo pipefail
+
+pkgcmd=${1:-"fedpkg"}
+branch=${2:-"rawhide"}
+run() {
+    local option=$1
+    local ext=$2
+    local expect=$3
+    \rm -rf noarch
```
richm, line 23: extra backslash?

check_rpmspec_collection.sh
nhosoi: > if there is more than one spec file you will get an error like this:
> 
> ```
> bash: [: another.spec: binary operator expected
> ```

Thanks for the solution. I was wondering if there were such case having multiple spec files (and hoping there isn't... :). But, there could be?

check_rpmspec_collection.sh
```diff
@@ -0,0 +1,52 @@
...
+    $cmdline > $output 2>1
+    count=$( ls noarch | wc -l )
+    if [ $count -eq $expect ]; then
+      echo OK - result $count is $expect
+    else
+      echo FAILED - result $count is not $expect
+    fi
+
+    filename=$( $pkgcmd --release $branch verrel )
+    rval=$( rpm -ql noarch/${filename}*.rpm | egrep '.html' | egrep collection | egrep -v '/usr/share/doc/' || : )
```
richm, line 36: you don't need `rpm -qlp` here?  I guess rpm is now smart enough to know that the given argument is the name of a package filename instead of a package.

richm: did you mean `fgrep` (literal match e.g. match the literal `.` in `.html` instead of treating `.` as "match any char") here instead of `egrep` (extended regular expression match)?

check_rpmspec_collection.sh
```diff
@@ -0,0 +1,52 @@
...
+    if [ "$rval" = "" ]; then
```
richm, line 37:
```
if rpm -ql noarch/${filename}*.rpm | egrep '.html' | egrep collection | egrep -v '/usr/share/doc/'; then
  echo FAILED - '.html' files are in noarch/${filename}*.rpm other than doc
else
  echo OK - '.html' files are only in doc in noarch/${filename}*.rpm
fi
```

check_rpmspec_collection.sh
richm: > But, there could be?

There should only ever be one file matching `*.spec` - so if you want to keep `if [ -f *.spec ]` that's fine - it should be pretty obvious to a developer what the problem is.

richm (approved):
lgtm

pcahyna:
What is the intended use of this tool? It seems that it makes plenty of restrictive assumptions about the build environment (only dist-git build supported).

nhosoi:
> What is the intended use of this tool? It seems that it makes plenty of restrictive assumptions about the build environment (only dist-git build supported).

That is correct. I'm going to use this script to verify the collection is properly packaged using the spec file. Luckily, the entire auto-maintenance is available in dist-git, I took advantage of it.

So far, it checks `--with collection_artifact` works as expected and README.html files are not included in the collection format. If necessary, e.g., we could add `galaxy-importer` to the test.

Issue #65, opened 2021-04-20
lsr_role2collection.py does not convert the legacy role path if the role is not the one to be converted.


nhosoi:
This is an example of README of sap-hana-preconfigure, in which some roles are referred.
```
./README.md:- 'linux-system-roles.timesync'
./README.md:- 'linux-system-roles.sap-base-settings' (for RHEL 7.x until RHEL 7.5)
./README.md:- 'linux-system-roles.sap-preconfigure' (for RHEL 7.6 and later and RHEL 8.x)
./README.md:        - { role: linux-system-roles.sap-base-settings }
./README.md:        - { role: linux-system-roles.sap-hana-preconfigure }
```
The corresponding converted result:
```
./roles/sap-hana-preconfigure/README.md:- 'linux-system-roles.timesync'
./roles/sap-hana-preconfigure/README.md:- 'linux-system-roles.sap-base-settings' (for RHEL 7.x until RHEL 7.5)
./roles/sap-hana-preconfigure/README.md:- 'linux-system-roles.sap-preconfigure' (for RHEL 7.6 and later and RHEL 8.x)
./roles/sap-hana-preconfigure/README.md:        - { role: linux-system-roles.sap-base-settings }
./roles/sap-hana-preconfigure/README.md:        - { role: sap.rhel.sap-hana-preconfigure }
```
All the roles must be converted to the collection format, but it'd be hard to automate it. The FQCNs need to be provided since, e.g., in this example, linux-system-roles.timesync could be either fedora.linux_system_roles.timesync or redhat.rhel_system_roles.timesync, and the others would be sap.rhel.sap-{base-settings,preconfigure}.

spetrosi:
@nhosoi was this issue resolved by #68?

nhosoi:
> @nhosoi was this issue resolved by #68?

Yep, thanks for the reminder, @spetrosi. Closing this issue.

PR #66, opened 2021-04-21
lsr_role2collection.py Introducing --new-role

spetrosi:
mssql and some other roles need to change this:
    ```
      - name: Run the role with default parameters
        include_role:
        name: linux-system-roles.mssql
    ```
    to this:
    ```
      - name: Run the role with default parameters
        include_role:
        name: microsoft.sql.server
    ```
    python lsr_role2collection.py --role mssql --new-role server --namespace microsoft --collection sql

PR #68, opened 2021-04-22
Fixing issue 65

nhosoi:
Fixing [lsr_role2collection.py does not convert the legacy role path if the role is not the one to be converted. ](https://github.com/linux-system-roles/auto-maintenance/issues/65)

lsr_role2collection.py - introducing --extra-roles EXTRA_ROLES
    
    The option takes comma separated extra roles to convert to collection.
    This pr fixes the extra roles belonging to the same namespace.collection
    as the role to be converted.
    
    Example:
    python  lsr_role2collection.py --role sap-hana-preconfigure \
      --extra-roles "sap-base-settings,sap-preconfigure"

lsr_role2collection.py - introducing --extra-mapping EXTRA_MAPPING
    
    Comma separated extra mapping - PRE:POST - to replace in the documentation
    
    The option takes comma separated extra string mapping (format PRE:POST)
    applied to the documentations. This option could be useful when a
    role is referred and the role does not belong to the same collection
    as the role to be converted.
    
    Example:
    This command line replaces "linux-system-roles.timesync" with
    "fedora.linux_system_roles.timesync" in the collection converted
    documentation.
    python lsr_role2collection.py --role sap-hana-preconfigure \
      --extra-roles "sap-base-settings,sap-preconfigure" \
      --namespace sap --collection rhel \
      --extra-mapping "linux-system-roles.timesync:fedora.linux_system_roles.timesync"

With these fixes, by running the above example command line, these roles in README.md
```
- 'linux-system-roles.timesync'
- 'linux-system-roles.sap-base-settings' (for RHEL 7.x until RHEL 7.5)
- 'linux-system-roles.sap-preconfigure' (for RHEL 7.6 and later and RHEL 8.x)
        - { role: linux-system-roles.sap-base-settings }
        - { role: linux-system-roles.sap-hana-preconfigure }
```
are converted to
```
- 'fedora.linux_system_roles.timesync'
- 'sap.rhel.sap-base-settings' (for RHEL 7.x until RHEL 7.5)
- 'sap.rhel.sap-preconfigure' (for RHEL 7.6 and later and RHEL 8.x)
          # rhel-system-roles.timesync variables
        - { role: sap.rhel.sap-base-settings }
        - { role: sap.rhel.sap-hana-preconfigure }
```
And these tasks
```
  tasks:
    - name: default run 0
      include_role:
        name: linux-system-roles.sap-preconfigure

    - name: default run 1
      include_role:
        name: linux-system-roles.sap-base-settings

    - name: default run 2
      include_role:
        name: linux-system-roles.sap-hana-preconfigure
```
are converted to
```  tasks:
    - name: default run 0
      include_role:
        name: sap.rhel.sap-preconfigure

    - name: default run 1
      include_role:
        name: sap.rhel.sap-base-settings

    - name: default run 2
      include_role:
        name: sap.rhel.sap-hana-preconfigure
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                        [--src-path SRC_PATH] [--src-owner SRC_OWNER] [--role ROLE]
                        [--new-role NEW_ROLE] [--replace-dot REPLACE_DOT]
                        [--subrole-prefix SUBROLE_PREFIX] [--readme README]
+                       README] [--extra-mapping EXTRA_MAPPING]
 ```
 
 ### optional arguments
@@ -172,6 +173,18 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                      If sub-role name does not start with the specified value, change
                      the name to start with the value; default to an empty string
 --readme README      Path to the readme file used in top README.md
+--extra-mapping EXTRA_MAPPING
+                     This is a comma delimited list of extra mappings to apply when converting the
+                     files - this converts the given name to collection format with the optional given
+                     namespace and collection.
+                     The format is
+                       "src_name:[[dest_namespace.]dest_collection.]dest_name,\
+                        src_name:[[dest_namespace.]dest_collection.]dest_name,...."
+                     The default for `dest_namespace` is the `--namespace` value,
+                     and the default for `dest_collection` is the `--collection` value.
+                     `src_name` is the name of a role, preferably in `namespace.rolename` format.
+                     If just using `rolename` for `src_name`, and `rolename` is used in places
+                     in the README that you do not want to change.
 ```
 
 ### environment variables
--- a/README.md
+++ b/README.md
@@ -184,7 +184,9 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                      and the default for `dest_collection` is the `--collection` value.
                      `src_name` is the name of a role, preferably in `namespace.rolename` format.
                      If just using `rolename` for `src_name`, and `rolename` is used in places
-                     in the README that you do not want to change.
+                     in the README that you do not want to change, you may have to change the
+                     README in another way, not using this script, by using sed with a custom
+                     regexp..
 ```
 
 ### environment variables
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                        [--src-path SRC_PATH] [--src-owner SRC_OWNER] [--role ROLE]
                        [--new-role NEW_ROLE] [--replace-dot REPLACE_DOT]
                        [--subrole-prefix SUBROLE_PREFIX] [--readme README]
-                       README] [--extra-mapping EXTRA_MAPPING]
+                       [--extra-mapping EXTRA_MAPPING]
 ```
 
 ### optional arguments
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                      If just using `rolename` for `src_name`, and `rolename` is used in places
                      in the README that you do not want to change, you may have to change the
                      README in another way, not using this script, by using sed with a custom
-                     regexp..
+                     regex.
 ```
 
 ### environment variables
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -565,6 +565,7 @@ class LSRFileTransformer(LSRFileTransformerBase):
                 if role[key].count(".") == 1:
                     _src_owner, _rolename_base = role[key].split(".")
                 else:
+                    _src_owner = None
                     _rolename_base = role[key]
                 if role[key] == lsr_rolename or self.comp_rolenames(
                     role[key], self.newrolename
@@ -573,30 +574,47 @@ class LSRFileTransformer(LSRFileTransformerBase):
                     changed = True
                 elif _rolename_base and _rolename_base in self.extra_mapping_src_role:
                     _src_role_index = self.extra_mapping_src_role.index(_rolename_base)
-                    role[key] = "{0}{1}".format(
-                        self.extra_mapping_dest_prefix[_src_role_index]
-                        if self.extra_mapping_dest_prefix[_src_role_index]
-                        else self.prefix,
-                        self.extra_mapping_dest_role[_src_role_index],
-                    )
-                    changed = True
+                    if (
+                        _src_owner is None
+                        or _src_owner
+                        and (
+                            _src_owner == self.extra_mapping_src_owner[_src_role_index]
+                            or _src_owner == self.src_owner
+                        )
+                    ):
+                        role[key] = "{0}{1}".format(
+                            self.extra_mapping_dest_prefix[_src_role_index]
+                            if self.extra_mapping_dest_prefix[_src_role_index]
+                            else self.prefix,
+                            self.extra_mapping_dest_role[_src_role_index],
+                        )
+                        changed = True
             else:
                 if role.count(".") == 1:
                     _src_owner, _rolename_base = role.split(".")
                 else:
+                    _src_owner = None
                     _rolename_base = role
                 if role == lsr_rolename or self.comp_rolenames(role, self.rolename):
                     role = self.prefix + self.newrolename
                     changed = True
                 elif _rolename_base and _rolename_base in self.extra_mapping_src_role:
                     _src_role_index = self.extra_mapping_src_role.index(_rolename_base)
-                    role = "{0}{1}".format(
-                        self.extra_mapping_dest_prefix[_src_role_index]
-                        if self.extra_mapping_dest_prefix[_src_role_index]
-                        else self.prefix,
-                        self.extra_mapping_dest_role[_src_role_index],
-                    )
-                    changed = True
+                    if (
+                        _src_owner is None
+                        or _src_owner
+                        and (
+                            _src_owner == self.extra_mapping_src_owner[_src_role_index]
+                            or _src_owner == self.src_owner
+                        )
+                    ):
+                        role = "{0}{1}".format(
+                            self.extra_mapping_dest_prefix[_src_role_index]
+                            if self.extra_mapping_dest_prefix[_src_role_index]
+                            else self.prefix,
+                            self.extra_mapping_dest_role[_src_role_index],
+                        )
+                        changed = True
             if changed:
                 item[roles_kw][idx] = role
 
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -411,7 +411,7 @@ class LSRFileTransformer(LSRFileTransformerBase):
             lsr_rolename = self.src_owner + "." + self.rolename
         logging.debug(f"\ttask role {rolename}")
         new_name = None
-        if rolename == self.rolename or rolename == lsr_rolename:
+        if rolename == lsr_rolename or self.comp_rolenames(rolename, self.rolename):
             new_name = self.prefix + self.newrolename
         elif _rolename_base and _rolename_base in self.extra_mapping_src_role:
             _src_role_index = self.extra_mapping_src_role.index(_rolename_base)
@@ -422,13 +422,9 @@ class LSRFileTransformer(LSRFileTransformerBase):
             # or current _rolename_base is SRC_ROLE1
             if (
                 not _src_owner
+                or (_src_owner == self.extra_mapping_src_owner[_src_role_index])
                 or (
-                    _src_owner
-                    and _src_owner == self.extra_mapping_src_owner[_src_role_index]
-                )
-                or (
-                    _src_owner
-                    and not self.extra_mapping_src_owner[_src_role_index]
+                    not self.extra_mapping_src_owner[_src_role_index]
                     and _src_owner == self.src_owner
                 )
             ):
@@ -554,10 +550,7 @@ class LSRFileTransformer(LSRFileTransformerBase):
         else:
             core0 = re.sub("[_\\.]", "", name0)
             core1 = re.sub("[_\\.]", "", name1)
-            if core0 == core1:
-                return True
-            else:
-                return False
+            return core0 == core1
 
     def change_roles(self, item, roles_kw):
         """ru_item is an item which may contain a roles or dependencies
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -422,8 +422,15 @@ class LSRFileTransformer(LSRFileTransformerBase):
             # or current _rolename_base is SRC_ROLE1
             if (
                 not _src_owner
-                or not self.extra_mapping_src_owner[_src_role_index]
-                or _src_owner == self.extra_mapping_src_owner[_src_role_index]
+                or (
+                    _src_owner
+                    and _src_owner == self.extra_mapping_src_owner[_src_role_index]
+                )
+                or (
+                    _src_owner
+                    and not self.extra_mapping_src_owner[_src_role_index]
+                    and _src_owner == self.src_owner
+                )
             ):
                 new_name = "{0}{1}".format(
                     self.extra_mapping_dest_prefix[_src_role_index]
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -130,6 +130,7 @@ class LSRFileTransformerBase(object):
         self.prefix = args["prefix"]
         self.subrole_prefix = args["subrole_prefix"]
         self.replace_dot = args["replace_dot"]
+        self.rolename_regex = "[{0}\\.]".format(self.replace_dot)
         self.role_modules = args["role_modules"]
         self.src_owner = args["src_owner"]
         self.top_dir = args["top_dir"]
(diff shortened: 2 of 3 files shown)


richm:
I get that we may need to fix naming and usage of roles from other roles.  So far we haven't had to deal with it in system roles, where some roles refer to other roles.  I guess `--extra-roles` implies that the roles are in the same namespace.collection, and `--extra-mapping` implies different namespace.collection?

richm:
I would prefer to have one option for extra mappings, that would work for any combination of namespace.collection.role or even namespace.collection.module - maybe something like 
```
--extra-mappings - this is a comma delimited list of extra mappings to apply when converting the files - this converts the given name to collection format with the given 
the format is

src_name:[[dest_namespace.]dest_collection.]dest_name,src_name:[[dest_namespace.]dest_collection.]dest_name,....

The default for `dest_namespace` is the `--namespace` value, and the default for `dest_collection` is the `--collection` value.

`src_name` is the name of a role, preferably in `namespace.rolename` format.  (We may in the future use `src_name` for modules and plugins).  If just using `rolename` for `src_name`, and `rolename` is used in places in the README that you do not want to change, then you may have to skip changing the README with this script and instead use `sed` and a regexp.
```

nhosoi:
> I would prefer to have one option for extra mappings, that would work for any combination of namespace.collection.role or even namespace.collection.module

Thank you for your inputs, @richm! Yes, let me update this pr as you suggested.

Indeed, I was hoping to have your comments on the design. When I looked into an other issue in sap, I happened to notice this bug which I did not anticipate. There could be more RFEs on the conversion tool as we support more roles such as sap and mssql... I'd like to hear from @spetrosi, as well. Thanks!

nhosoi:
@richm, @spetrosi,
> I would prefer to have one option for extra mappings, that would work for any combination of namespace.collection.role or even namespace.collection.module - maybe something like
>
> ```
> --extra-mappings - this is a comma delimited list of extra mappings to apply when converting the files -
> this converts the given name to collection format with the optional given namespace and collection.
>
> the format is
>
> src_name:[[dest_namespace.]dest_collection.]dest_name,src_name:[[dest_namespace.]dest_collection.]dest_name,....
>
> The default for `dest_namespace` is the `--namespace` value,
> and the default for `dest_collection` is the `--collection` value.
>
> `src_name` is the name of a role, preferably in `namespace.rolename` format.
> (We may in the future use `src_name` for modules and plugins).
> If just using `rolename` for `src_name`, and `rolename` is used in places in the
> README that you do not want to change, then you may have to skip changing the
> README with this script and instead use `sed` and a regexp.
> ```

Let me double-check what you meant here...

When we are trying to convert `sap-hana-preconfigure` with `--role sap-hana-preconfigure`, we pass
--extra-mapping "sap-base-settings" instead of --extra-mapping "linux-system-roles.sap-base-settings:microsoft.rhel.sap-base-settings"not to change `- { role: linux-system-roles.sap-base-settings }` in README.md?
But if we don't want to replace `linux-system-roles.sap-base-settings`, we can make it happen by not setting `linux-system-roles.sap-base-settings` with `--extra-mapping` at all, I'd think.
Or you meant it's just about the role `sap-hana-preconfigure` itself? I.e., if we set `--extra-mapping "sap-hana-preconfigure"`, even if we are converting `sap-hana-preconfigure`, the role name in README will not be converted?

I have another question.
When converting mssql `--role mssql` with --namespace microsoft, --collection sql ,
if we set `--extra-mapping "linux-system-roles.mssql:server"`, it'd result in the same outcome as from [lsr_role2collection.py Introducing --new-role #66](https://github.com/linux-system-roles/auto-maintenance/pull/66).
Do we want to close pr66 in favor of this spec? Or it's ok to keep both of them?

In my original thought for --extra-roles and --extra-mapping, I did not plan to allow the current role (option of --role) in --extra-roles and --extra-mapping. But now we are going to allow?

richm:
> When we are trying to convert sap-hana-preconfigure with --role sap-hana-preconfigure, we pass
--extra-mapping "sap-base-settings" instead of --extra-mapping "linux-system-roles.sap-base-settings:microsoft.rhel.sap-base-settings"not to change - { role: linux-system-roles.sap-base-settings } in README.md?

If you are using `lsr_role2collection.py --namespace sap --collection rhel --role sap-hana-preconfigure --new-role hana_preconfigure` then you would use `--extra-mappings linux-system-roles.sap-base-settings:base_settings` - this would look for occurrences of `linux-system-roles.sap-base-settings` and change them to `sap.rhel.base_settings`.  The `src_name` is `linux-system-roles.sap-base-settings` - the `dest_namespace` is from `--namespace sap` and the `dest_collection` is from `--collection rhel` - the `dest_name` is `base_settings` - so the FQCN dest name is `sap.rhel.base_settings`.  This would change
`- { role: linux-system-roles.sap-base-settings }` to `- { role: sap.rhel.base_settings }` in README.md.

> I have another question.
> When converting mssql --role mssql with --namespace microsoft, --collection sql ,
> if we set --extra-mapping "linux-system-roles.mssql:server", it'd result in the same outcome as from lsr_role2collection.py Introducing --new-role #66.
> Do we want to close pr66 in favor of this spec? Or it's ok to keep both of them?

Hmm - good question - I was thinking that `--new-role` would be for the role being converted, and `--extra-mappings` would be for references to roles other than the role being converted.  But as you point out, `--extra-mappings` could be used for all of those cases.  In fact, we could have a single flag - `--mappings` - which would replace `--namespace`, `--collection`, `--role`, and `--new-role`, and would also handle "extra" mappings as well.  But I think it is not necessary to make a change like that to the api - I think we should have both`--new-role` and `--extra-mappings`, and revisit this issue if it becomes a problem.  I think having explicit flags for namespace, collection, role, new-role, etc. is more clear.

> In my original thought for --extra-roles and --extra-mapping, I did not plan to allow the current role (option of --role) in --extra-roles and --extra-mapping. But now we are going to allow?

No, I don't think so.

nhosoi:
This pr has been updated.

lsr_role2collection.py - introducing --extra-mapping EXTRA_MAPPING
    
    EXTRA_MAPPING: This is a comma delimited list of extra mappings to
        apply when converting the files - this converts the given name to
        collection format with the optional given namespace and collection.
        The format is "src_name:[[dest_namespace.]dest_collection.]dest_name,\
                       src_name:[[dest_namespace.]dest_collection.]dest_name,...."
        The default for `dest_namespace` is the `--namespace` value, and the
        default for `dest_collection` is the `--collection` value. `src_name`
        is the name of a role, preferably in `namespace.rolename` format.
        If just using `rolename` for `src_name`, and `rolename` is used in
        places in the README that you do not want to change.
    
    Example:
    $ python lsr_role2collection.py --role sap-hana-preconfigure \
        --extra-mapping "linux-system-roles.sap-preconfigure:ns0.col0.preconfigure,\
        sap-base-settings:col1.base-settings"
    
    Legacy format:
      roles:
        - linux-system-roles.sap-preconfigure
        - linux-system-roles.sap-base-settings
        - linux-system-roles.sap-hana-preconfigure
    
    Converted collection format:
      roles:
        - ns0.col0.preconfigure
        - fedora.col1.base-settings
        - fedora.linux_system_roles.sap-hana-preconfigure

README.md
```diff
@@ -171,6 +171,18 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
...
+                     files - this converts the given name to collection format with the optional given
+                     namespace and collection.
+                     The format is
+                       "src_name:[[dest_namespace.]dest_collection.]dest_name,\
+                        src_name:[[dest_namespace.]dest_collection.]dest_name,...."
+                     The default for `dest_namespace` is the `--namespace` value,
+                     and the default for `dest_collection` is the `--collection` value.
+                     `src_name` is the name of a role, preferably in `namespace.rolename` format.
+                     If just using `rolename` for `src_name`, and `rolename` is used in places
+                     in the README that you do not want to change.
```
richm, line 185: "do not want to change, you may have to change the README in another way, not using this script, by using `sed` with a custom regexp."

richm: looking at the code - looks like `src_name` is treated as `src_owner.src_rolename`?  Does this mean if the user specifies a `src_name` like `storage` with no `.`, does the code look for `linux-system-roles.storage` and try to convert that to `fedora.system_roles.storage`?  Does it look for both `storage` and `linux-system-roles.storage`?

nhosoi: This is an example yml:
```
  roles:
    - name: linux-system-roles.storage
    - name: selinux
    - name: rhel-system-roles.logging
```
These 3 command lines converts the example yml to the same format [0].
`$ python  lsr_role2collection.py --role testrole --extra-mapping "storage:storage,selinux:selinux,rhel-system-roles.logging:logging"`
`$ python  lsr_role2collection.py --role testrole --extra-mapping "storage:storage,linux-system-roles.selinux:selinux,rhel-system-roles.logging:logging"`
`$ python  lsr_role2collection.py --role testrole --extra-mapping "linux-system-roles.storage:storage,linux-system-roles.selinux:selinux,rhel-system-roles.logging:logging"
[0]
```
  roles:
    - name: fedora.linux_system_roles.storage
    - name: fedora.linux_system_roles.selinux
    - name: fedora.linux_system_roles.logging
```

richm: > This is an example yml:
> 
> ```
>   roles:
>     - name: linux-system-roles.storage
>     - name: selinux
>     - name: rhel-system-roles.logging
> ```
> 
> These 3 command lines converts the example yml to the same format [0].
> `$ python lsr_role2collection.py --role testrole --extra-mapping "storage:storage,selinux:selinux,rhel-system-roles.logging:logging"`
> `$ python lsr_role2collection.py --role testrole --extra-mapping "storage:storage,linux-system-roles.selinux:selinux,rhel-system-roles.logging:logging"`
> `$ python lsr_role2collection.py --role testrole --extra-mapping "linux-system-roles.storage:storage,linux-system-roles.selinux:selinux,rhel-system-roles.logging:logging"
> [0]
> 
> ```
>   roles:
>     - name: fedora.linux_system_roles.storage
>     - name: fedora.linux_system_roles.selinux
>     - name: fedora.linux_system_roles.logging
> ```

What if you had this?
```yaml
 roles:
    - name: storage
    - name: linux-system-roles.storage
    - name: selinux
    - name: rhel-system-roles.logging
```
?

nhosoi: Ok. There is a bug. :) This is converted to `- name: fedora.linux_system_roles.storage`, as well.
```
  roles:
    - name: something-else.storage
```

richm:
There is a lot of code duplication here.  There are at least 3 places in the code which do something like this:
```
            if rolename.count(".") == 1:
                _src_owner, _rolename_base = rolename.split(".")
            else:
                _src_owner = None
                _rolename_base = rolename
```
https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L415
https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L559
https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L587

and this:

https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L431-L441
https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L571-L584
https://github.com/linux-system-roles/auto-maintenance/blob/d46e2123888cc419f63d4af8517910f75a2351dd/lsr_role2collection.py#L597-L610

nhosoi:
> There is a lot of code duplication here. There are at least 3 places in the code which do something like this:
> 
> 8< -------------------------------------------- >8

Yes, I had to repeat these codes to handle the mapping cases... Do you recommend to make them a function (macro?)? Could there be any good way to do it in python? Or you think the duplication is unnecessary?

richm:
> > There is a lot of code duplication here. There are at least 3 places in the code which do something like this:
> > 8< -------------------------------------------- >8
> 
> Yes, I had to repeat these codes to handle the mapping cases... Do you recommend to make them a function

Yes

>  (macro?)?

I don't think python has macros.

> Could there be any good way to do it in python? Or you think the duplication is unnecessary?

I think the duplication is not necessary.

nhosoi:
> > Could there be any good way to do it in python? Or you think the duplication is unnecessary?
> 
> I think the duplication is not necessary.

Could you please give me some more details how to do it? Such as storing the info somewhere and reuse it in the next time type of modification?

richm:
> > > Could there be any good way to do it in python? Or you think the duplication is unnecessary?
> > 
> > 
> > I think the duplication is not necessary.
> 
> Could you please give me some more details how to do it? Such as storing the info somewhere and reuse it in the next time type of modification?

I submitted a pull request against your branch

nhosoi:
> I submitted a pull request against your branch

Thanks a lot, @richm. Merged your pr and fixed my logic bug. It passes all of my tests now.

lsr_role2collection.py
```diff
@@ -389,6 +394,56 @@ class LSRFileTransformer(LSRFileTransformerBase):
...
+        if rolename.count(".") == 1:
+            _src_owner, _rolename_base = rolename.split(".")
+        else:
+            _src_owner = None
+            _rolename_base = rolename
+        if not lsr_rolename:
+            lsr_rolename = self.src_owner + "." + self.rolename
+        logging.debug(f"\ttask role {rolename}")
+        new_name = None
+        if rolename == self.rolename or rolename == lsr_rolename:
```
richm, line 408: The previous code did this when being called from `task_cb`:
```
            if rolename == self.rolename or rolename == lsr_rolename:
```
but this when being called from `change_roles`:
```
                if role[key] == lsr_rolename or self.comp_rolenames(
                    role[key], self.rolename
                ):
```
Do you know why the comparison is different in the `task_cb` case and the `change_roles` case?  Why is `comp_rolenames` used?

richm: not sure - but I submitted another PR against your branch that uses `comp_rolenames` and cleans up some logic

nhosoi:
> Do you know why the comparison is different in the task_cb case and the change_roles case? Why is comp_rolenames used?

I'm also not sure why, but `comp_rolenames` is needed when the role name contains `.` or `_` (we had to replace `.` with `_` to make ansible happier). And except metrics, we did not have to deal with it. But I'd think it's safer to use `comp_rolenames` all the time. That said, if a role contains '.' and it's replaced with an option --replace-dot, the value should be included in `core = re.sub("[_\\.]", "", name)` in `comp_rolenames` to be ignored in the comparison... Is it ok to fix it in this bz? (I think it's a minor issue, though...)

> not sure - but I submitted another PR against your branch that uses comp_rolenames and cleans up some logic

Thank you, @richm. Merged. (thank you for cleaning up the redundancy in the logic! :p)

nhosoi:
> > Do you know why the comparison is different in the task_cb case and the change_roles case? Why is comp_rolenames used?
> 
> I'm also not sure why, but `comp_rolenames` is needed when the role name contains `.` or `_` (we had to replace `.` with `_` to make ansible happier). And except metrics, we did not have to deal with it. But I'd think it's safer to use `comp_rolenames` all the time. That said, if a role contains '.' and it's replaced with an option --replace-dot, the value should be included in `core = re.sub("[_\\.]", "", name)` in `comp_rolenames` to be ignored in the comparison... Is it ok to fix it in this bz? (I think it's a minor issue, though...)

Done.

README.md
```diff
@@ -141,6 +141,7 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                        [--src-path SRC_PATH] [--src-owner SRC_OWNER] [--role ROLE]
                        [--new-role NEW_ROLE] [--replace-dot REPLACE_DOT]
                        [--subrole-prefix SUBROLE_PREFIX] [--readme README]
+                       README] [--extra-mapping EXTRA_MAPPING]
```
richm, line 144: typo?

lsr_role2collection.py
```diff
@@ -1214,6 +1341,21 @@ def process_readme(
         dest = roles_dir / new_role
         file_patterns = ["*.md"]
         file_replace(dest, src_owner + "." + role, prefix + new_role, file_patterns)
+        # --extra-mapping SRCROLENAME:DESTROLENAME(or FQDN)
```
richm, line 1344: FQCN

README.md
```diff
@@ -172,6 +173,20 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
...
+                     The format is
+                       "src_name:[[dest_namespace.]dest_collection.]dest_name,\
+                        src_name:[[dest_namespace.]dest_collection.]dest_name,...."
+                     The default for `dest_namespace` is the `--namespace` value,
+                     and the default for `dest_collection` is the `--collection` value.
+                     `src_name` is the name of a role, preferably in `namespace.rolename` format.
+                     If just using `rolename` for `src_name`, and `rolename` is used in places
+                     in the README that you do not want to change, you may have to change the
+                     README in another way, not using this script, by using sed with a custom
+                     regexp..
```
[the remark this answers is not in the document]
nhosoi: Sorry, fixed...

richm:
lgtm - please squash

richm (approved):
lgtm

richm:
> lgtm - please squash

or, just use the "Squash and merge" feature of the Merge pull request button.

nhosoi:
> > lgtm - please squash
> 
> or, just use the "Squash and merge" feature of the Merge pull request button.

Done. Thanks a lot, @richm.

PR #69, opened 2021-04-28
Improve dependencies in the packaging

nhosoi:
collection_requirements.txt - added
collection_readme.md - make dependencies programmable.

### DIFF
--- a/lsr_role2collection/collection_readme.md
+++ b/lsr_role2collection/collection_readme.md
@@ -13,8 +13,8 @@ NOTE: Some roles are not supported in RHEL6 and RHEL7. For more details about th
 
 ## Dependencies
 
-The following dependency is required for the Ansible Controller:
-* jmespath
+If installing from RPM, the dependencies will be installed with the package.
+Otherwise, the dependencies are listed in `requirements.txt` and/or `bindep.txt`.
 
 ## Installation
 
--- a/lsr_role2collection/collection_requirements.txt
+++ b/lsr_role2collection/collection_requirements.txt
@@ -1,0 +1,2 @@
+jmespath
+netaddr
--- a/release_collection.py
+++ b/release_collection.py
@@ -41,6 +41,12 @@ def build_collection(src_path, dest_path, galaxy, coll_rel, force):
     ignore_file_dir = os.path.join(coll_dir, "tests", "sanity")
     ignore_file = os.path.join(ignore_file_dir, "ignore-2.9.txt")
     collection_readme = os.path.join("lsr_role2collection", "collection_readme.md")
+    collection_requirements = os.path.join(
+        "lsr_role2collection", "collection_requirements.txt"
+    )
+    collection_requirements_dest = os.path.join(coll_dir, "requirements.txt")
+    collection_bindep = os.path.join("lsr_role2collection", "collection_bindep.txt")
+    collection_bindep_dest = os.path.join(coll_dir, "bindep.txt")
     ignore_file_src = os.path.join("lsr_role2collection", "extra-ignore-2.9.txt")
     ansible_lint = os.path.join("lsr_role2collection", ".ansible-lint")
     for rolename, roledata in coll_rel.items():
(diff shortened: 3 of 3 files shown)


richm:
Can you give a usage example?

pcahyna:
> Can you give a usage example?

I would be interested in that too. It is not clear where will the information come from.

nhosoi:
> > Can you give a usage example?
> 
> I would be interested in that too. It is not clear where will the information come from.

This dist-git pr/7 is the beneficiary of this pr/69.
https://src.fedoraproject.org/fork/nhosoi/rpms/linux-system-roles/c/baeafb4778304fc58bbd83f5ff52b2426a392bf3

I was hoping to update the dist-git pr/7, once this pr is merged. But I guess it'd be better to discuss on the both pr's to make them in sync...

richm:
I don't think we should be trying to parse the python requirements format into individual names and operations in the spec file, then put them back together in the correct python format in gen-requirements.py. I think the `--requirements` argument value should already be in the correct python requirements.txt format.  I also think it should be in the canonical form, which is one package per line:
```
ansible ; python_version > "2.6"
ansible<2.7 ; python_version < "2.7"
idna<2.8 ; python_version < "2.7"
PyYAML<5.1 ; python_version < "2.7"
```

For upstream, we should have a requirements.txt file in the `lsr_role2collection/` subdirectory.  The `galaxy_transform.py` script could update the `%dependencies%` in `collection_readme.md` with the contents of `requirements.txt`.

For downstream - the spec file should create a `requirements.txt` file in the correct python format.

nhosoi:
collection_readme.md - make dependencies programmable.
collection_requirements.txt - added

richm:
Ok.  We'll need this in the Galaxy collection as well, so please update https://github.com/linux-system-roles/auto-maintenance/blob/master/release_collection.py to add these two files as well, and update the collection README.md

lsr_role2collection/collection_readme.md
```diff
@@ -13,8 +13,8 @@ NOTE: Some roles are not supported in RHEL6 and RHEL7. For more details about th
 
 ## Dependencies
 
-The following dependency is required for the Ansible Controller:
-* jmespath
+The following dependencies are required for the Ansible Controller:
+%dependencies%
```
richm, line 17: After thinking about this some more - I don't think we need a `Dependencies` section.

If the user is installing from Galaxy or Automation Hub, the user should know to look in `requirements.txt` and/or `bindep.txt` if the tooling doesn't use those files automatically (e.g. if you do `ansible-galaxy collection install fedora.system_roles` does it know to look in `requirements.txt` and ensure those dependencies?  If not, then it is the responsibility of the user).

If installing from RPM, the dependencies should automatically be installed - in that case, we should tell the user that.

Perhaps this wording:
```
## Dependencies

If installing from RPM, the dependencies will be installed with the package.  Otherwise, the dependencies are listed
in `requirements.txt` and/or `bindep.txt`.
```

nhosoi:
> Ok. We'll need this in the Galaxy collection as well, so please update https://github.com/linux-system-roles/auto-maintenance/blob/master/release_collection.py to add these two files as well, and update the collection README.md

I'll follow your [next suggestion](https://github.com/linux-system-roles/auto-maintenance/pull/69#pullrequestreview-652877402). So, I guess it's not an issue any more. But out of curiosity, to upload to Galaxy, don't we (or can't we) use the collection artifact, e.g., fedora-linux_system_roles-1.1.0.tar.gz? It can be built using the same way in the linux-system-roles dist-git. In creating the package, we could update the README.md for the both fedora and redhat here https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/7#_2__14-26.

richm:
> > Ok. We'll need this in the Galaxy collection as well, so please update https://github.com/linux-system-roles/auto-maintenance/blob/master/release_collection.py to add these two files as well, and update the collection README.md
> 
> I'll follow your [next suggestion](https://github.com/linux-system-roles/auto-maintenance/pull/69#pullrequestreview-652877402). So, I guess it's not an issue any more. But out of curiosity, to upload to Galaxy, don't we (or can't we) use the collection artifact, e.g., fedora-linux_system_roles-1.1.0.tar.gz?

We can, but why would we go through the trouble of building a Fedora rpm just to get the collection bits out of it to upload to Galaxy?

I suppose at some point in the future when we can do:
- github merge to a role repo triggers a Fedora event
- the Fedora event updates the Fedora spec file with the latest role commit hash
- the Fedora event triggers an RPM build using the role commit hash
- the Fedora event extracts the collection bits out and publishes to Galaxy

but it might be easier to have a github action that triggers on the github role repo merge to build and publish the collection

> It can be built using the same way in the linux-system-roles dist-git. In creating the package, we could update the README.md for the both fedora and redhat here https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/7#_2__14-26.

richm (dismissed):
lgtm

nhosoi:
> lgtm

Sorry, @richm. I had to fix the black error (;_;). Could you approve one more time? Thank you!!

richm (approved):
lgtm

PR #70, opened 2021-05-19
lsr_role2collection.py - adding an ability to convert "- ROLENAME" to "- FQCN" in the doc files such as README

nhosoi:
Note: Stop skipping the src role is the role to be converted to support the
case - the converted role name is changed and the original role name
appears in a doc file.


richm (approved):
lgtm

nhosoi:
Thanks for your review, @richm!

PR #71, opened 2021-05-25
Substitute asciidoc with asciidoctor. Check if asciidoctor is present.

spetrosi:


pcahyna:
Do we have asciidoctor in the RHEL versions that we build against?

spetrosi:
Forget about it, `rubygem-asciidoctor` is not available for RHEL 7 and 8 so we must use `asciidoc`.

PR #72, opened 2021-05-25
 Fix formatting issue with redundant plus signes when converting with pandoc

spetrosi:
Add check for the asciidoc package installed and update error message

 Fix formatting issue with redundant plus signes

### DIFF
--- a/md2html.sh
+++ b/md2html.sh
@@ -35,6 +35,9 @@ for file in "$@"; do
   # With pandoc, convert to adoc, then to HTML
   elif [ "$md2html_tool" == pandoc ]; then
     $md2html_tool -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+    # Fix formatting issue with redundant plus signes
+    sed -r -i -e "s/\`\+([^\`]*)\+\`/\`\1\`/g" "${file%.md}.tmp.adoc"
+
     touch -r "${file}" "${file%.md}.tmp.adoc"
     TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
     tr -d '\r' < "${file%.md}.html" > "${file%.md}.tmp.adoc"
--- a/md2html.sh
+++ b/md2html.sh
@@ -36,7 +36,9 @@ for file in "$@"; do
   elif [ "$md2html_tool" == pandoc ]; then
     $md2html_tool -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
     # Fix formatting issue with redundant plus signes
-    sed -r -i -e "s/\`\+([^\`]*)\+\`/\`\1\`/g" "${file%.md}.tmp.adoc"
+    # Disable shellcheck because single quotes are used intentionally
+    # shellcheck disable=SC2016
+    sed -i -e 's/`+\([^`]*\)+`/`\1`/g' "${file%.md}.tmp.adoc"
 
     touch -r "${file}" "${file%.md}.tmp.adoc"
     TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
--- a/md2html.sh
+++ b/md2html.sh
@@ -24,7 +24,7 @@ for file in "$@"; do
   # RHEL 9 in brew cannot use pandoc, hence trying kramown first
   if type -p kramdown >/dev/null && kramdown -i GFM < /dev/null > /dev/null; then
     md2html_tool=kramdown
-  elif type -p pandoc >/dev/null; then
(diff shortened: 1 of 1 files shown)


md2html.sh
```diff
@@ -35,13 +35,16 @@ for file in "$@"; do
   # With pandoc, convert to adoc, then to HTML
   elif [ "$md2html_tool" == pandoc ]; then
     $md2html_tool -f markdown_github "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
+    # Fix formatting issue with redundant plus signes
+    sed -r -i -e 's/`\+([^`]*)\+`/`\1`/g' "${file%.md}.tmp.adoc"
```
pcahyna, line 39: do you know what is the root cause? sed tricks are very fragile, so it would be preferable to solve the issue in a different way, if possible.

spetrosi: No idea what the root cause is. I only know that this is caused by pandoc. I blame the gfm converter being unstable as gfm is not as widely used as markdown. I was not able to find any info on the internet neither.

spetrosi:
Shellcheck finds unrelated issues with the sed expression and I needed to ignore the checks:
```
In md2html.sh line 39:
    sed -r -i -e "s/`\+([^\`]*)\+`/`\1`/g" "${file%.md}.tmp.adoc"
                    ^------------^ SC2006: Use $(...) notation instead of legacy backticked `...`.
                       ^-- SC1036: '(' is invalid here. Did you forget to escape it?
                       ^-- SC1088: Parsing stopped here. Invalid use of parentheses?
                                   ^--^ SC2006: Use $(...) notation instead of legacy backticked `...`.
```

spetrosi:
Weirdly, this issue only happens for me locally, the READMEs in the RPM are fine
https://koji.fedoraproject.org/koji/buildinfo?buildID=1732985

pcahyna:
> Shellcheck finds unrelated issues with the sed expression and I needed to ignore the checks:
> 
> ```
> In md2html.sh line 39:
>     sed -r -i -e "s/`\+([^\`]*)\+`/`\1`/g" "${file%.md}.tmp.adoc"
>                     ^------------^ SC2006: Use $(...) notation instead of legacy backticked `...`.
>                        ^-- SC1036: '(' is invalid here. Did you forget to escape it?
>                        ^-- SC1088: Parsing stopped here. Invalid use of parentheses?
>                                    ^--^ SC2006: Use $(...) notation instead of legacy backticked `...`.
> 
> Did you mean: 
>     sed -r -i -e "s/$(\+([^\`]*)\+)/$(\1)/g" "${file%.md}.tmp.adoc"
> ```

Are you sure it is unrelated? It is better to understand the problem and fix it instead of ignoring checks.

spetrosi:
Understood the problem and fixed it instead of ignoring checks as per Pavel's suggestion by escaping backticks :)

pcahyna:
Well you could have solved the backtick problem in a more readable way by using apostrophes for quoting the whole string instead of double quotes.

richm:
> Well you could have solved the backtick problem in a more readable way by using apostrophes for quoting the whole string instead of double quotes.

e.g.
```
sed -e 's/`+\([^`]*\)+`/`\1`/g'
```

spetrosi:
shellcheck didn't like single quotes:
```
In md2html.sh line 39:
    sed -i -e 's/`+\([^`]*\)+`/`\1`/g' "${file%.md}.tmp.adoc"
              ^----------------------^ SC2016: Expressions don't expand in single quotes, use double quotes for that.

For more information:
  https://www.shellcheck.net/wiki/SC2016 -- Expressions don't expand in singl...
ERROR: InvocationError for command /usr/bin/bash -c 'shellcheck -e SC1090 *.sh' (exited with code 1)
```

Do you prefer ignoring this shellcheck issue and using single quotes for readability?

richm:
> https://www.shellcheck.net/wiki/SC2016

I guess when you have backticks inside of quotes, shellcheck thinks you are trying to use the shell backtick command expansion?  Seems like an overly aggressive check - I would prefer that it assumes the user knows what she/he is doing when the user uses single quotes.  So I would prefer to use single quotes and ignore this shellcheck.

pcahyna:
> > https://www.shellcheck.net/wiki/SC2016
> 
> I guess when you have backticks inside of quotes, shellcheck thinks you are trying to use the shell backtick command expansion? Seems like an overly aggressive check - I would prefer that it assumes the user knows what she/he is doing when the user uses single quotes. So I would prefer to use single quotes and ignore this shellcheck.

Indeed. shellcheck uses heuristics to guess what you might have wanted, and in this case it does not take into account the possibility that you actually wanted a literal backtick inside a string. Therefore, one can not blindly follow what shellchecks recommends, nor blindly ignore the warning. One has to understand the reason for the warning first.

richm:
ready for review?

spetrosi:
No review, apparently the issue with the redundant plus signs only happened to me locally. In brew builds the formatting is correct.
I am not able to reproduce this locally now neither after upgrading to F34, no idea what the issue was.
I am closing this PR as it's NOTABUG.

PR #74, opened 2021-06-02
md2html.sh - avoid using GFM extension

nhosoi:
- If the md file uses gfm style code blocks, convert them to depth 4
  indent.
- For the kramdown input option, instead of GFM, use kramdown.

### DIFF
--- a/md2html.sh
+++ b/md2html.sh
@@ -19,30 +19,57 @@ while getopts "hl" opt; do
   esac
 done
 
+md2html_tool=""
+# RHEL 9 in brew cannot use pandoc, hence trying kramdown first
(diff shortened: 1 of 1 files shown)


spetrosi:
I am afraid that the difference between kramdown and gfm is not only the codeblock syntax. I tried this md2html.sh on the logging role and the resulted HTML renders many other things incorrectly.
We may convert all READMEs to markdown in the repositories using pandoc for example, then use kramdown to convert md>HTML during builds.

nhosoi:
As rubygem-kramdown-parser-gfm is resurrected, we can continue using GFM. Hence, closing this pr.

PR #78, opened 2021-06-08
lsr_role2collection.py - Adding encoding="utf-8" to open.

nhosoi:
Some version fails to open a file containing non ascii characters, e.g., `author: First Škarvada <fskarvad@example.com>` as follows:
```
  File "lsr_role2collection.py", line 281, in run
    filepath, self.role_name, self.transformer_args
  File "lsr_role2collection.py", line 136, in __init__
    buf = open(filepath).read()
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc5 in position 32: ordinal not in range(128)
```
To avoid the failures, adding `encoding="utf-8"` to open.


richm (approved):
lgtm

PR #79, opened 2021-06-08
lsr_role2collection.py - dropping python-six dependency

nhosoi:
lsr_role2collection.py - drop python-six dependency
    
Instead of using "six.string_types" to check the task's module name,
use simple "str". If the script is run by python2 and the module name
is not ascii string, we may get unexpected results. But the chance
should be minimum.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -35,7 +35,6 @@ import textwrap
 from pathlib import Path
 from ruamel.yaml import YAML
 from shutil import copytree, copy2, copyfile, ignore_patterns, rmtree
-from six import string_types
 from operator import itemgetter
 
 ALL_ROLE_DIRS = [
@@ -505,7 +504,7 @@ class LSRFileTransformer(LSRFileTransformerBase):
                         _match.group(2),
                     )
             elif (
-                isinstance(task[module_name], string_types)
+                isinstance(task[module_name], str)
                 and _src_owner_match in task[module_name]
             ):
                 _path = task[module_name]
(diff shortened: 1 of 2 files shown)


lsr_role2collection.py
```diff
@@ -447,6 +446,12 @@ def convert_rolename(self, rolename, lsr_rolename=None):
                 )
         return new_name
 
+    def isstring(self, input):
+        try:
+            return isinstance(input, basename)
```
richm, line 451: basestring?

nhosoi: Thank you for catching it! (I don't know why my test did not capture it... I guess to test it I should have a non-ascii task name...)

nhosoi: Anyway, I've fixed my typo... :p

richm: > @richm, Thank you for catching it! ~(I don't know why my test did not capture it... I guess to test it I should have a non-ascii task name...)~
> 
> Reconsidering this change, I'm forcing to run lsr_role2collection.py with python3 even in the older env like rhel-7. That's said, we could simply replace the line 513 with `isinstance(task[module_name], str)`?

Yes, good idea.

> 
> Shebang of this script is `#!/usr/bin/env python3`. Could there be anything else we could claim this script only runs with python3?

nhosoi: Thank you, @richm. Updated.

richm (approved):
lgtm

spetrosi (approved):
I've reviewed the changes and they look good to me. However, I am far not an expert and have no confidence in my review..

nhosoi:
Thank you for your reviews, @richm and @spetrosi. Merging...

PR #80, opened 2021-06-15
pandoc - use gfm instead of markdown

richm:
pandoc should use `gfm` instead of `markdown_github`.  This makes the
internal links in the HTML work when converting the resulting
asciidoc to HTML.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1962976


nhosoi (approved):
lgtm
Thanks for the fix, @richm!

PR #83, opened 2021-07-08
release 1.6.0; add meta/runtime.yml to collection

richm:
Ansible Galaxy collection import now requires that the collection
has a meta/runtime.yml.  This commit adds support for a default
meta/runtime.yml, and allows the user to override this when using
lsr_role2collection.

Other changes for 1.6.0 release


nhosoi (approved):
lgtm

Thanks for adding the file.

PR #92, opened 2021-08-10
report-modules-plugins.py - static analysis of Ansible plugins

richm:
This script will analyze one or more Ansible roles and collections
looking for plugins which are built-in or not, and produce a report.


richm:
For example, this is what the report looks like for the logging role:
```
The following ansible.builtin modules are used at runtime:
command copy debug fail file include_role include_tasks include_vars lineinfile package package_facts service set_fact setup stat systemd template
The following ansible.builtin filters are used at runtime:
basename bool comment flatten intersect ternary to_nice_json union
The following ansible.builtin tests are used at runtime:
file match version
The following ansible.builtin lookups are used at runtime:
lookup
The following jinja2 filters are used at runtime:
d default format int join length list map rejectattr selectattr trim
The following jinja2 tests are used at runtime:
defined mapping sequence string undefined

The following additional plugins are used at runtime:
community.general.seport type: module
	role: logging.rsyslog
		roles/rsyslog/tasks/inputs/ovirt/main.yml:28


The following ansible.builtin modules are used in testing:
assert command copy debug fail file find import_role include_role include_tasks meta package package_facts set_fact shell stat
The following ansible.builtin filters are used in testing:
basename difference product
The following ansible.builtin tests are used in testing:
match
The following jinja2 filters are used in testing:
int join list map
The following jinja2 tests are used in testing:
defined

The following additional plugins are used in testing:
community.crypto.openssl_certificate type: module
	role: logging
		tests/tasks/create_tests_certs.yml:32
		tests/tasks/create_tests_certs.yml:53
community.crypto.openssl_csr type: module
	role: logging
		tests/tasks/create_tests_certs.yml:26
		tests/tasks/create_tests_certs.yml:44
community.crypto.openssl_privatekey type: module
	role: logging
		tests/tasks/create_tests_certs.yml:21
		tests/tasks/create_tests_certs.yml:39
```

tyll:
Not sure why but it seems to be missing instances of json_query like these: https://github.com/linux-system-roles/network/blob/main/defaults/main.yml#L31

nhosoi:
I happened to have this file community/general/tests/integration/targets/inventory_docker_machine/playbooks/test_inventory_1.yml in my .ansible/collections and this task [0] crashes report-modules-plugins.py when trying to parse {{.Name}} | {{.DriverName}} | {{.State}} | {{.URL}} | {{.Error}}` at [1].
[0]
```
- hosts: 127.0.0.1
  gather_facts: no
  tasks:
  - name: sanity check Docker Machine output
    vars:
      dm_ls_format: !unsafe '{{.Name}} | {{.DriverName}} | {{.State}} | {{.URL}} | {{.Error}}'
      success_regex: "^vm | [^|]+ | Running | tcp://.+ |$"
    command: docker-machine ls --format '{{ dm_ls_format }}'
    register: result
    failed_when: result.rc != 0 or result.stdout is not match(success_regex
)
```
[1]
> /path/to/linux-system-roles/auto-maintenance/report-modules-plugins.py(295)find_plugins()
-> tmpl = filectx.templar.environment.parse(source=args)

The task does not belong to us. And I'm not sure this is a valid value (at least yamllint does not complain)... Do you think it's a syntax to be supported by the parser?

richm:
> The task does not belong to us. And I'm not sure this is a valid value (at least yamllint does not complain)... Do you think it's a syntax to be supported by the parser?

No.  That's golang template format.  Hmm - will need a special exception for that one.

richm:
> > The task does not belong to us. And I'm not sure this is a valid value (at least yamllint does not complain)... Do you think it's a syntax to be supported by the parser?
> 
> No. That's golang template format. Hmm - will need a special exception for that one.

Fixed.

richm:
If the string cannot be parsed as a Jinja2 template, it will emit a warning like this:
```
WARNING:root:the string [{{.Name}} | {{.DriverName}} | {{.State}} | {{.URL}} | {{.Error}}] could not be processed as a Jinja2 template at /path/to/file.yml:1
```

richm:
I'd like to merge this, as other groups are starting to use this and want to contribute patches.  Any objections?

PR #97, opened 2021-08-24
combine collection release and update into release_collection.py

richm:
The release collection script can now automatically update
collection_release.yml and galaxy.yml, and can publish the
new collection to Galaxy.


nhosoi:
lgtm

I have a request, which is out of the scope of this pr and it may be an overkill..., but could it be possible to extend this tool to accept an rpm package as an input to generate the collection release package?

richm:
> lgtm
> 
> I have a request, which is out of the scope of this pr and it may be an overkill..., but could it be possible to extend this tool to accept an rpm package as an input to generate the collection release package?

Yeah, I think that would be useful for a follow-up PR.  I can think of two use cases:
1) given an RPM, use this script to extract the collection data to use with `ansible-galaxy collection build` and build the collection
2) Use this script inside the rpm spec to assemble the collection instead of the shell commands we are currently using e.g. the code to remove the .dotfiles, assemble the ignore files, etc.

nhosoi:
> > lgtm
> > I have a request, which is out of the scope of this pr and it may be an overkill..., but could it be possible to extend this tool to accept an rpm package as an input to generate the collection release package?
> 
> Yeah, I think that would be useful for a follow-up PR. I can think of two use cases:
> 
>     1. given an RPM, use this script to extract the collection data to use with `ansible-galaxy collection build` and build the collection
> 
>     2. Use this script inside the rpm spec to assemble the collection instead of the shell commands we are currently using e.g. the code to remove the .dotfiles, assemble the ignore files, etc.

I wish 2 is as straightforward as it should be, the thing is ansible (& ansible-galaxy) is not always available in the spec file... So, I'm hoping to have the use case 1 more.

nhosoi (approved):

PR #103, opened 2021-10-14
release_collection.py - Add --skip-git option; Skip ansible-galaxy if it does not exist.

nhosoi:
Planning to use release_collection.py in the spec file with the --skip-git option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,6 +473,7 @@ will publish the collection, waiting until it is completed.
 * `--no-wait` - boolean - By default, when publishing, the script will wait
   until the publishing is completed.  Use `--no-wait` if you do not want to
   wait, and instead will check the import status in Galaxy.
+* `--skip-git` - boolean - If set to `true`, use local source. By default, `false`.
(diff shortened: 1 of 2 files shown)


nhosoi:
> ok - so how will this work from inside the spec file? Will the spec file create a file like `collection_release.yml` based on each rolename and `%deftag` or `%defcommit`? I guess that is only needed for the list of roles to include - which can also be provided by the `--include` argument.

This is my testing branch.
https://src.fedoraproject.org/fork/nhosoi/rpms/linux-system-roles/blob/release_collection/f/linux-system-roles.spec

I'm running release_collection.py with the list of included roles like this.
```
declare -A includes
for role in %{rolenames}; do
    includes="$includes --include $role"
done

python3 release_collection.py --galaxy-yml galaxy.yml --src-path $(pwd) \
    --dest-path $(pwd)/.collections $includes --force --skip-git --no-update
``
We can simplify the spec file a lot with this replacement.`

nhosoi:
Thanks, @richm!

PR #113, opened 2021-12-10
Add initial automate-collections-updates.py

spetrosi:


spetrosi:
Hi @richm and @nhosoi,

I am pushing here the script that I wrote for automating collections updates based on what Rich wrote in 
[update_and_return_versions.py ](https://github.com/linux-system-roles/ee_linux_system_roles/blob/main/.github/scripts/update_and_return_versions.py).
This is my first python script that makes some actual enterprise work, please be patient :) I will be happy for any feedback and will do my best to apply it properly.

spetrosi:
The script works for me locally, and the next step would be to deploy it somewhere. The system must have access to rhpkg or brew commands, I do not know how to implement this now. I am testing the script by running it with my rhpkg. Any ideas?

nhosoi:
@spetrosi, looks very useful!

I do not have an answer to your auth question, but I'd believe this pr/113 (well, the subset of this pr :) should be able to replace this code https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/38# in the spec file(s). Don't you think it's cleaner to reduce automate-collections-updates.py to fit in the spec file, eliminate the pr/38 change from the spec file and use automate-collections-updates.py in the build?

automate-collections-updates/automate_collections_updates.py
```diff
@@ -0,0 +1,154 @@
...
+        help="The repository to work on. Either rhel-system-roles or linux-system-roles",
+    )
+    args = parser.parse_args()
+
+    hsh = yaml.safe_load(open(args.vendored_collections))
+    collection_tarballs = {}
+
+    repo = args.repository
+    if repo == "rhel-system-roles":
+        repo_url = "https://src.osci.redhat.com:443/rpms/rhel-system-roles.git"
```
richm, line 36: should definitely not have this url in a public file - I would expect these settings will be provided in some sort of config file.  and the internal config file will only be stored internally somewhere

automate-collections-updates/automate_collections_updates.py
```diff
@@ -0,0 +1,154 @@
...
+                    "--single-branch",
+                ]
+            )
+            """Move tarballs to the repo directory"""
+            for tarball in collection_tarballs.values():
+                shutil.move(os.path.abspath(tarball), os.path.join(repo, tarball))
+            """Upload new sources"""
+            os.chdir(repo)
+            subprocess.run(
+                [pkg_command, "upload", " ".join(collection_tarballs.values())]
```
richm, line 131: We only upload to centos stream or rhel - we don't use vendoring on Fedora.

automate-collections-updates/automate_collections_updates.py
```diff
@@ -0,0 +1,154 @@
...
+        coll_dir = os.path.join(
+            coll_install_dir,
+            "ansible_collections",
+            coll_name.split(".")[0],
+            coll_name.split(".")[1],
+        )
+        cmd = [
+            "ansible-galaxy",
+            "collection",
+            "install",
```
richm, line 64: Can you use `download` instead of `install`, to directly download a .tar.gz file?  Or is it that the logic that figures out what is the latest release only works with `install` and not `download`?

automate-collections-updates/automate_collections_updates.py
```diff
@@ -0,0 +1,154 @@
...
+                stdout=subprocess.PIPE,
+                encoding="utf-8",
+            )
+            pat = "Task info: (.*$)"
+            build_url = re.search(pat, out.stdout).group(1)
+            print(f"Build info: {build_url}")
+            # TODO: Share build_url somewhere
+        finally:
+            print("Cleaning up the downloaded repository")
+            os.chdir("..")
```
richm, line 149: which directory is this script supposed to be run from?  I think you'll have to be careful about paths if you assume a certain directory, or don't assume a cwd anywhere - all of the `open` calls should construct the full path and filename - use the `cwd` argument of `subprocess.run` to run the command in a certain directory.

automate-collections-updates/automate_collections_updates.py
```diff
@@ -0,0 +1,154 @@
...
+            pat = f"Installing '{coll_name}:([^']+)' to"
+            coll_latest_ver = re.search(pat, out.stdout).group(1)
+            if coll_latest_ver == coll["version"]:
+                print(f"{coll_name} is of latest version. Skipping.")
+            else:
+                print(f"Searching {coll_name} changelog for updates.")
+                for value in coll_content_values:
+                    changelog = yaml.safe_load(
+                        open(os.path.join(coll_dir, "changelogs/changelog.yaml"))
+                    )
```
richm, line 85: do we need to look for updates in the changelog?  Can't we just assume if the version is different, it is newer, and use it?  I guess that would cause problems if they updated the version a lot for trivial, non-API issues, but I don't think that is the case.

richm:
> @spetrosi, looks very useful!
> 
> I do not have an answer to your auth question, but I'd believe this pr/113 (well, the subset of this pr :) should be able to replace this code https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/38# in the spec file(s). Don't you think it's cleaner to reduce automate-collections-updates.py to fit in the spec file, eliminate the pr/38 change from the spec file and use automate-collections-updates.py in the build?

No, I don't think the specfile should be pulling collection tarballs from galaxy and automation hub. I think that should be a separate process.  Pulling sources in the specfile is against the RPM way, which is to list them in the `sources` file and store them internally in the lookaside cache.

richm:
> The script works for me locally, and the next step would be to deploy it somewhere. The system must have access to rhpkg or brew commands, I do not know how to implement this now. I am testing the script by running it with my rhpkg. Any ideas?

For centos, there is the `centpkg` command, which you can create a token and store it in the ~/.config/rpkg/centpkg.conf file.  That might be good enough.  We can store that token (or even the entire centpkg.conf) internally in bitwarden.  It should definitely not be made public.

automate-collections-updates/automate_collections_updates.py
spetrosi: `ansible-galaxy collection` cannot download tar.gz only. The best it can do is the following
`-n, --no-deps         Don't download collections listed as dependencies.`
So the script needs to install the collection and then tar it. Luckily, the format of the collection installed through `ansible-galaxy` is the same as in the tar downloaded from AH via UI.

automate-collections-updates/automate_collections_updates.py
spetrosi: Good point, I'll fix this

automate-collections-updates/automate_collections_updates.py
spetrosi: I pushed this as a suggestion for you to look at. And now with your comment I understand that it is a loss of time to check for the changelog. It is not a big deal for us to vendor a new version if it got released. I'll remove this logic.

automate-collections-updates/automate_collections_updates.py
richm: ?  I'm using this logic in another script and it works fine:
`ansible-galaxy collection download -vv -p . ansible.posix`
e.g.
```
Starting collection download process to '/home/rmeggins/rh-dist-git/tests/rhel-system-roles/Library/basic/ansible.posix'
Downloading https://galaxy.ansible.com/download/ansible-posix-1.3.0.tar.gz to /home/rmeggins/.ansible/tmp/ansible-local-36821042yavan3/tmpz01gfpll/ansible-posix-1.3.0-tmwjrmic
Downloading collection 'ansible.posix:1.3.0' to '/home/rmeggins/rh-dist-git/tests/rhel-system-roles/Library/basic/ansible.posix'
Collection 'ansible.posix:1.3.0' was downloaded successfully
Writing requirements.yml file of downloaded collections to '/home/rmeggins/rh-dist-git/tests/rhel-system-roles/Library/basic/ansible.posix/requirements.yml'
```
this creates the file `ansible-posix-1.3.0.tar.gz`

spetrosi: Yuy, it does work with `ansible-galaxy` from `ansible-core`, I'll use it here.

richm:
lgtm - just the formatting issue to fix

richm (approved):
lgtm

PR #121, opened 2022-01-26
[WIP] Improve update_vendored_collections.py and prepare to run in container

spetrosi:
Please review for Python best practices


spetrosi:
Note to self: At the end of the script, update [auto-maintenance/vendored_collections.yml](https://github.com/linux-system-roles/auto-maintenance/blob/master/update_vendored_collections/vendored_collections.yml) with new current versions and push here to open a PR

richm (approved):
lgtm

richm:
I don't think the collection tarballs should be committed to git

spetrosi:
> I don't think the collection tarballs should be committed to git

You are right, we cannot publicly share AH tarballs. Thank you for the input, removed

richm (dismissed):
lgtm
probably the version of black changed recently - we should fix that in a separate commit

richm (approved):
lgtm

PR #127, opened 2022-02-05
add support for --check-fqcn

richm:

PR #128, opened 2022-02-05
add support for --extra-script for custom conversions

richm:


README.md
```diff
@@ -190,6 +190,13 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                      in the README that you do not want to change, you may have to change the
                      README in another way, not using this script, by using sed with a custom
                      regex.
+--extra-script /path/to/executable
+                     This is a script to use to do custom conversion of the role.  For example,
+                     to convert uses of filter plugins to FQCN.  If you do not specify anything,
+                     then it looks for an executable file named `lsr_role2coll_extra_script` in
+                     the role root directory, and runs it.  The arguments such as dest dir,
+                     namespace, etc. are passed in via environment variables.  See the code for
+                     the list of environment variables available.
```
nhosoi, line 199: Do we already have an implementation of `lsr_role2coll_extra_script` to convert filter plugins to FQCN? (I'd like to try it in my test env... :)

richm: > Do we already have an implementation of `lsr_role2coll_extra_script` to convert filter plugins to FQCN? (I'd like to try it in my test env... :)

https://github.com/linux-system-roles/vpn/pull/46

Note that this is just a `sed` - the problem with filter plugins is that there is no way to extract the exact position of the filter plugin in a jinja template - unlike modules, which can easily be replaced - so any implementation of lsr_role2collection which tried to replace all types of plugins would have to use some sort of regexp search/replace, which brings a whole host of corner cases and other problems.  The lsr_role2collection script can't possibly handle all of these cases without a lot of help from the role developer.

So instead, `--extra-script` puts that burden on the role developer - the role developer is responsible for creating and writing the plugins in such a way that it is easy to replace them with the FQCN, and is responsible for providing the implementation.

https://github.com/linux-system-roles/auto-maintenance/pull/127 makes it easy for role developers to identify where the plugins are used, so that they can ensure their `lsr_role2coll_extra_script` will find and fix all of the local plugins that need to be converted to FQCN.

nhosoi: +1

Prior to your commet, I tested with this simple sed and it worked like a charm. After running lsr_role2collections with --extra-script, #127 returned "Found 0 errors"!
```
FILES="tasks/mesh_conf.yml \
       templates/libreswan-host-to-host.conf.j2 \
       templates/libreswan-host-to-host.secrets.j2"
for file in $FILES; do
    sed -i -e "s/\(vpn_ipaddr\)/redhat.rhel_system_roles.\1/g" $file
done
```
If `report-modules-plugins.py --check-fqcn` did return lots of items which should be FQDN, I might have asked to add another option to `report-modules-plugins.py` to generate a script for each role to convert the failed item to FQDN, but so far there is just one. So, I guess it's not that useful at all... :)

nhosoi (approved):
lgtm

spetrosi (approved):
lgtm

PR #149, opened 2022-05-04
support collection-requirements.yml

richm:
There is no way for a role to automatically require collections for
module/plugin dependencies.  Use the non-standard file
`meta/collection-requirements.yml` for this.  Note that
`meta/requirements.yml` is supposed to be a list of roles only, in
the format described at https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#installing-multiple-roles-from-multiple-files
If we find a `meta/requirements.yml` file, and it is not in `list`
format, assume it is an old one that has not yet been renamed to
`meta/collection-requirements.yml`, use it, and warn the user.

### DIFF
--- a/release_collection.py
+++ b/release_collection.py
@@ -340,18 +340,30 @@ def update_collection(args, galaxy, coll_rel):
             galaxy["name"],
             collection_readme,
         )
-        req_yml = os.path.join(args.src_path, rolename, "meta", "requirements.yml")
-        if os.path.isfile(req_yml):
-            req_yml_hsh = yaml.safe_load(open(req_yml))
-            for coll in req_yml_hsh.get("collections", []):
-                if isinstance(coll, dict):
-                    coll_name = coll["name"]
-                    coll_ver = coll.get("version", "*")
-                else:
-                    coll_name = coll
-                    coll_ver = "*"
-                galaxy_deps = galaxy.setdefault("dependencies", {})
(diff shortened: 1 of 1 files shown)


richm:
This is related to https://github.com/linux-system-roles/tox-lsr/pull/78

release_collection.py
```diff
@@ -340,18 +340,30 @@ def update_collection(args, galaxy, coll_rel):
...
+        coll_rqf = "collection-requirements.yml"
+        for rqf in [legacy_rqf, coll_rqf]:
+            req_yml = os.path.join(args.src_path, rolename, "meta", rqf)
+            if os.path.isfile(req_yml):
+                req_yml_hsh = yaml.safe_load(open(req_yml))
+                if isinstance(req_yml_hsh, list):
+                    continue  # legacy role format
+                if rqf == legacy_rqf:
+                    logging.warning(
+                        "The role %s is still using %s - please convert to %s instead",
```
nhosoi, line 353: I'm wondering to convert `requirements.yml` to `collection-requirements.yml`, are there anything else to do other than renaming the filename?

nhosoi, line 353: I'm wondering to convert `requirements.yml` to `collection-requirements.yml`, are there anything else to do other than renaming the filename?

richm: No.

nhosoi (approved):
lgtm

PR #155, opened 2022-06-01
lsr_role2collection.py - add FQCN substitution to support "Roles Used by Other Roles"

nhosoi:
In addition to `linux-system-roles.rolename:fedora.linux_system_roles.rolename`,
`--extra-mapping` takes `fedora.linux_system_roles.rolename:redhat.rhel_system_roles.rolename`.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -1086,8 +1086,9 @@ def role2collection():
         default=os.environ.get("COLLECTION_EXTRA_MAPPING", ""),
         help=(
             "This is a comma delimited list of extra mappings to apply when "
-            "converting the files - this converts the given name to collection "
-            "format with the optional given namespace and collection."
+            "converting the files - this replaces the given role name with "
+            "collection format with the optional given namespace and collection "
+            "as sell as the given FQCN with other FQCN."
         ),
     )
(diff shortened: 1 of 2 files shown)


lsr_role2collection.py
```diff
@@ -1615,6 +1665,27 @@ def process_readme(
             env=env,
         )
 
+    # Handle --extra-mapping FQCN0:FQCN1 or FQCN2:ROLE3
+    file_patterns = ["*.yml", "*.md", "*.py"]
```
richm, line 1669: `.py` files?

nhosoi: I thought there could be some slightest possibility, but no such thing... :p
If so, we don't have to check files under the plugin_dir.
https://github.com/linux-system-roles/auto-maintenance/pull/155/files#diff-f5898ea097abc3a298c8aa8850d649247348992056a612047ca43828e0b7a56aR1682-R1683

nhosoi: I agree with you. :) I'm removing them...

richm (dismissed):
lgtm

richm:
looks like you need to rebase to the latest master, but otherwise, please merge

PR #162, opened 2022-06-30
roles-tag-and-release.sh - edit CHANGELOG.md

nhosoi:
When editing CHANGELOG.md, if the file starts with "# Changelog" or
"Changelog/n===...", keep them there.

### DIFF
--- a/roles-tag-and-release.sh
+++ b/roles-tag-and-release.sh
@@ -103,12 +103,12 @@ if [ "$skip" = false ]; then
         read -r -p "Edit release notes - press Enter to continue"
         rel_notes_file=".release-notes-${new_tag}"
         if [ ! -f "$rel_notes_file" ]; then
-            echo "[$new_tag] - $( date +%Y-%m-%d )" > "$rel_notes_file"
-            echo "--------------------" >> "$rel_notes_file"
-            echo "" >> "$rel_notes_file"
-            echo "REMOVE_ME: Recommend to itemize the change logs in either of the following two." >> "$rel_notes_file"
-            echo "### New features" >> "$rel_notes_file"
-            echo "### Bug fixes" >> "$rel_notes_file"
+            { echo "[$new_tag] - $( date +%Y-%m-%d )"; \
+              echo "--------------------"; \
+              echo ""; \
+              echo "REMOVE_ME: Recommend to itemize the change logs in either of the following two."; \
+              echo "### New features"; \
+              echo "### Bug fixes"; } > "$rel_notes_file"
             git log --oneline --no-merges --reverse --pretty=format:"- %B" "${latest_tag}".. | \
                 tr -d '\r' >> "$rel_notes_file"
         fi
@@ -119,8 +119,7 @@ if [ "$skip" = false ]; then
 ========="
             clheader=$(head -2 CHANGELOG.md)
             if [ "$myheader" = "$clheader" ]; then
-                echo "$clheader" > .tmp-changelog
-                echo "" >> .tmp-changelog
+                { echo "$clheader"; echo ""; } > .tmp-changelog
                 cat "$rel_notes_file" >> .tmp-changelog
                 tail -n +3 CHANGELOG.md >> .tmp-changelog
             else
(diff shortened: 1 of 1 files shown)


roles-tag-and-release.sh
```diff
@@ -106,7 +106,21 @@ if [ "$skip" = false ]; then
         ${EDITOR:-vi} "$rel_notes_file"
         if [ -f CHANGELOG.md ]; then
             read -r -p "Edit CHANGELOG.md - press Enter to continue"
-            cat "$rel_notes_file" CHANGELOG.md > .tmp-changelog
+            pat="=*"
+            if [[ $( head -n 1 CHANGELOG.md ) == "# Changelog" ]]; then
```
richm, line 110: Why do we have two different styles of changelog file?  We should have only 1

nhosoi: If we don't have to support
```
Changelog
=========
```
I'm more than happy! I'm eliminating it and just support `# Changelog`.
Thanks!

richm: note that the network role is the only role that currently has a CHANGELOG and it uses
```
Changelog
=========
```
What if we only use that form?  Is there somewhere else that uses the `# Changelog` form?

nhosoi: Well, I already generated the initial CHANGELOG.md for most of the roles using `github-changelog-generator`, which generates `# Changelog` form...

richm: ok - I think the network role shouldn't mind changing the `====` form to the `# Changelog` form - either way is fine

nhosoi: Fingers crossed... May I ask them if they don't mind?

richm: > Fingers crossed... May I ask them if they don't mind?

Sure

roles-tag-and-release.sh
```diff
@@ -106,7 +106,17 @@ if [ "$skip" = false ]; then
         ${EDITOR:-vi} "$rel_notes_file"
         if [ -f CHANGELOG.md ]; then
             read -r -p "Edit CHANGELOG.md - press Enter to continue"
-            cat "$rel_notes_file" CHANGELOG.md > .tmp-changelog
+			pat="=*"
```
richm, line 109:
```bash
  myheader="Changelog
========="
  clheader=$(head -2 CHANGELOG.md)
  if [ "$myheader" = "$clheader" ]; then
    echo "$clheader" > .tmp-changelog
    cat "$rel_notes_file" >> .tmp-changelog
    tail -n +3 CHANGELOG.md >> .tmp-changelog
  else
    echo WARNING: Changelog header "$clheader"
    echo not in expected format "$myheader"
    cat "$rel_notes_file" CHANGELOG.md > .tmp-changelog
  fi
```

nhosoi: Thanks, @richm. If we are more conscious about the markdown format, `roles-tag-and-release.sh` generates `## [VERSION] - DATE` and stores in `$rel_notes_file`. We still don't know to run `markdownlint.yml` action in each role, but considering the rules set for the network role, we'd like to have `[1.9.4] - 2022-06-28` in the Setext style. Do you think it's possible? (please note that it's ok to have `### New features` and `### Bug fixes`.)

Example of `$rel_notes_file`.
```
## [1.9.4] - 2022-06-28

### New features

- use gather_facts: true in tests_include_vars_from_parent
- support ansible-core-2.13

### Bug fixes

- make min_ansible_version a string in meta/main.yml
```

nhosoi: BTW, it's not a good idea to replace the ATX format with Setext before inserting "$rel_notes_file" like this?
```
sed -i -e "s/^# Changelog/Changelog\n=========/" \
       -e "s/^## \(.*\)/\1\n--------------------/" CHANGELOG.md
```
Is it too brutal?

richm (approved):
lgtm

nhosoi:
> lgtm

Thank you, @richm. I made a small change to support `v##.##.##` style for `sshd`.

roles-tag-and-release.sh
```diff
@@ -56,9 +57,13 @@ fi
...
+# special case for network and sshd
 case "$latest_tag" in
-v*) latest_tag="${latest_tag//v}" ;;
+v*)
+    if [ "$repo" != "ansible-sshd" ]; then
+        latest_tag="${latest_tag//v}"
+    fi
+    ;;
 esac
 skip=false
```
richm, line 68: add
```
# shellcheck disable=SC2129
```
to make shellcheck tests pass

richm (approved):
lgtm

PR #165, opened 2022-07-10
use role-make-version-changelog.sh instead of roles-tag-and-release.sh

richm:
Use role-make-version-changelog.sh instead of roles-tag-and-release.sh
because of the new version/tag/release workflow centered around
CHANGELOG.md.

Use `gh` instead of `hub`.


role-make-version-changelog.sh
```diff
@@ -0,0 +1,198 @@
...
+        # sshd uses a leading v
+        allow_v=v
+    else
+        # network had a case where there where two tags for the
+        # latest - one with leading v and one without - git describe
+        # would return the one with the leading v - so had to strip
+        # it off
+        latest_tag="${latest_tag//v}"
+    fi
+    ;;
```
nhosoi, line 100: If the branch is not tagged with any version, `git describe` fails as follows.
```
$ git describe --tags --abbrev=0
fatal: No names found, cannot describe anything.
$ echo $?
128
```
Could it be possible to add another question asking if an initial CHANGELOG.md file to be generated or not?
If the answer is `yes`, can we add new CHANGELOG.md having the contents something like this? (Maybe, we could ask another question what initial version to be set? And the date will be auto-filled...)
```
Changelog
=========

[#.#.#] - YYYY-MM-DD
--------------------

# Initial Release
```

richm: Fixed.  Latest commit works if no tags and will allow you to create the CHANGELOG.md

nhosoi (approved):
Thank you for the updates, @richm!

lgtm

PR #171, opened 2022-07-22
Update release_collection.py

nhosoi:
- Instead of distutils.version use pkg_resources.
- Fix a bug in compact_coll_changelog not to include a line starting
  with '-' in an example code, e.g.,
```
  ```yaml
  - set_fact:
      var: "{{ var + item }}"
      loop: "{{ some_list }}"
```

### DIFF
--- a/release_collection.py
+++ b/release_collection.py
@@ -29,7 +29,7 @@ except ImportError:
 
 import json
 
-from distutils.version import StrictVersion
+from pkg_resources import parse_version
 
 DEFAULT_GIT_SITE = "https://github.com"
 DEFAULT_GIT_ORG = "linux-system-roles"
@@ -61,6 +61,17 @@ def run_cmd(cmdlist, cwd=None, env=None):
     return rc
 
 
+def lsr_parse_version(v_str):
+    test_v = parse_version("1")  # guaranteed to work
+    v = parse_version(v_str)
+    if not isinstance(v, test_v.__class__):
+        raise ValueError(f"Error: {v_str} is not a proper version number")
+    if not hasattr(v, "release"):
+        if hasattr(v, "_version"):
+            setattr(v, "release", v._version.release)
+    return v
+
+
 def check_versions_updated(cur_ref, new_ref, versions_updated):
     """Compare versions and update list.
 
@@ -69,10 +80,10 @@ def check_versions_updated(cur_ref, new_ref, versions_updated):
     minor, and/or micro versions were updated."""
 
     try:
-        cur_v = StrictVersion(cur_ref)
-        new_v = StrictVersion(new_ref)
+        cur_v = lsr_parse_version(cur_ref)
+        new_v = lsr_parse_version(new_ref)
         for idx in range(0, 3):
-            if cur_v.version[idx] != new_v.version[idx]:
+            if cur_v.release[idx] != new_v.release[idx]:
                 versions_updated[idx] = True
     except ValueError as exc:
         logging.debug(f"Could not compare version {cur_ref} to {new_ref}: {exc}")
@@ -84,25 +95,19 @@ def comp_versions(cur_ref, new_ref):
     """Compare versions.
 
     Return values:
-    1  - if cur_ref > new_ref
-    0  - if cur_ref == new_ref
-    -1 - if cur_ref < new_ref"""
+    -1 - if cur_ref < new_ref
+     0 - if cur_ref == new_ref
+     1 - if cur_ref > new_ref"""
 
     try:
-        _mobj = re.match(r"v?(\d*[.]\d*[.]\d*)", cur_ref)
-        if _mobj and _mobj.group(1):
-            cur_ref = _mobj.group(1)
-        _mobj = re.match(r"v?(\d*[.]\d*[.]\d*)", new_ref)
-        if _mobj and _mobj.group(1):
-            new_ref = _mobj.group(1)
-        cur_v = StrictVersion(cur_ref)
-        new_v = StrictVersion(new_ref)
-        for idx in range(0, 3):
-            if cur_v.version[idx] > new_v.version[idx]:
-                return 1
-            elif cur_v.version[idx] < new_v.version[idx]:
-                return -1
-        return 0
+        v_cur_ref = lsr_parse_version(cur_ref)
+        v_new_ref = lsr_parse_version(new_ref)
+        if v_cur_ref < v_new_ref:
+            return -1
+        elif v_cur_ref == v_new_ref:
+            return 0
+        else:
+            return 1
     except ValueError as exc:
         logging.debug(f"Could not compare version {cur_ref} to {new_ref}: {exc}")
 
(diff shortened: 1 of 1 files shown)


richm:
I think we do not have to worry about python2 support.  Everywhere we want to run, there is python 3.6 at least and python3-setuptools.
Note that `parse_version` will accept a string that begins with a `v`:
```
>>> v1_str = 'v1.2.3'
>>> v2_str = '3.2.1'
>>> v1 = pkg_resources.parse_version(v1_str)
>>> v2 = pkg_resources.parse_version(v2_str)
>>> v1
<Version('1.2.3')>
>>> v2
<Version('3.2.1')>
>>> v1 < v2
True
```
There are a couple of big problems with `parse_version`.  The first is that it doesn't throw an error if the version is bogus.
```
>>> v3 = pkg_resources.parse_version("asdfasdfasdf")
>>> v3
<LegacyVersion('asdfasdfasdf')>
>>> v3.release
>>> 
```
I'm not sure how to check if parsing failed.  Here is the implementation used in python 3.6: https://github.com/pypa/setuptools/blob/v39.2.0/pkg_resources/__init__.py#L117
```python
def parse_version(v):
    try:
        return packaging.version.Version(v)
    except packaging.version.InvalidVersion:
        return packaging.version.LegacyVersion(v)
```
The simple implementation would be to say `if isinstance(ver, LegacyVersion): raise` - the problem is that with 3.6, the `packaging` class is vendored into `pkg_resources`, and that module takes great pains to make the Version classes hidden and private.  I'm not sure how to import it.

Another problem - the python 3.6 `Version` class has no `release` method or member.  It does have `_version.release` though.  So the code in `check_versions_updated` will have to do something like this
```python
        cur_v = parse_version(cur_ref)
        new_v = parse_version(new_ref)
        try:
            cur_release = cur_v.release
        except AttributeError:
            cur_release = cur_v._version.release
```

I think we should have our own `lsr_parse_version` function.
```python
def lsr_parse_version(v_str):
    test_v = parse_version("1")  # guaranteed to work
    v = parse_version(v_str)
    if not isinstance(v, test_v.__class__):
        raise ValueError(f"Error: {v_str} is not a proper version number")
    if not hasattr(v, 'release'):
        if hasattr(v, '_version'):
            setattr(v, 'release', v._version.release)
    return v
```
that way, we are guaranteed that
* an exception is raised if the version could not be parsed
* the version object returned has a `release` member that is a tuple of the version numbers

nhosoi:
On 7/22/22 15:02, Richard Megginson wrote:
>
> I think we do not have to worry about python2 support.
>
Ok. That makes the code a bit cleaner... I used 
python-setuptools-0.9.8-7.el7.noarch to test on rhel-7.9 (actually, it 
is CentOS Linux release 7.9.2009, though), but indeed no need to stick 
with python2...

And I made sure we haven't ported the code using release_collection.py 
to the extra-rhel-7.9 spec file. It still directly uses 
lsr_role2collection. I guess we don't have to change the part to upgrade 
the roles themselves. So, I'm stopping to support the older setuptools.
>
> Everywhere we want to run, there is python 3.6 at least and 
> python3-setuptools.
> Note that |parse_version| will accept a string that begins with a |v|:
>
> |>>> v1_str = 'v1.2.3' >>> v2_str = '3.2.1' >>> v1 = 
> pkg_resources.parse_version(v1_str) >>> v2 = 
> pkg_resources.parse_version(v2_str) >>> v1 <Version('1.2.3')> >>> v2 
> <Version('3.2.1')> >>> v1 < v2 True |
>
> There are a couple of big problems with |parse_version|. The first is 
> that it doesn't throw an error if the version is bogus.
>
> |>>> v3 = pkg_resources.parse_version("asdfasdfasdf") >>> v3 
> <LegacyVersion('asdfasdfasdf')> >>> v3.release >>> |
>
> I'm not sure how to check if parsing failed. Here is the 
> implementation used in python 3.6: 
> https://github.com/pypa/setuptools/blob/v39.2.0/pkg_resources/__init__.py#L117
>
> def  parse_version(v):
>      try:
>          return  packaging.version.Version(v)
>      except  packaging.version.InvalidVersion:
>          return  packaging.version.LegacyVersion(v)
>
> The simple implementation would be to say |if isinstance(ver, 
> LegacyVersion): raise| - the problem is that with 3.6, the |packaging| 
> class is vendored into |pkg_resources|, and that module takes great 
> pains to make the Version classes hidden and private. I'm not sure how 
> to import it.
>
> Another problem - the python 3.6 |Version| class has no |release| 
> method or member. It does have |_version.release| though. So the code 
> in |check_versions_updated| will have to do something like this
>
>          cur_v  =  parse_version(cur_ref)
>          new_v  =  parse_version(new_ref)
>          try:
>              cur_release  =  cur_v.release
>          except  AttributeError:
>              cur_release  =  cur_v._version.release
>
> I think we should have our own |lsr_parse_version| function.
>
> def  lsr_parse_version(v_str):
>      test_v  =  parse_version("1")# guaranteed to work
>      v  =  parse_version(v_str)
>      if  not  isinstance(v,test_v.__class__):
>          raise  ValueError(f"Error: {v_str} is not a proper version number")
>      if  not  hasattr(v,'release'):
>          if  hasattr(v,'_version'):
>              setattr(v,'release',v._version.release)
>      return  v
>
> that way, we are guaranteed that
>
>   * an exception is raised if the version could not be parsed
>   * the version object returned has a |release| member that is a tuple
>     of the version numbers
>
All right. Let me work on lsr_parse_version.

Thanks a lot, Rich!!
--noriko

nhosoi:
Thank you for your advice, @richm. I updated the pr as you suggested and squashed into one commit.

release_collection.py
```diff
@@ -280,8 +285,10 @@ def update_galaxy_version(args, galaxy, versions_updated):
         and any(versions_updated[0:3])
         and not args.no_auto_version
     ):
-        galaxy_ver = StrictVersion(galaxy["version"])
-        major, minor, micro = galaxy_ver.version
+        galaxy_ver = lsr_parse_version(galaxy["version"])
+        major = galaxy_ver.major
```
richm, line 289:
```python
         major, minor, micro = galaxy_ver.release
```

nhosoi: Sorry, I missed it... (I tested with python3.6, but there was no problem reported... (;_;) I wish it failed there...):

richm: > Sorry, I missed it... (I tested with python3.6, but there was no problem reported... (;_;) I wish it failed there...):

How did you test with python 3.6?  When I test, I'm doing something like this:
```
podman run --rm -it centos:7 bash
yum -y install python3-setuptools
python3
from pkg_resources import parse_version
....
```

richm (approved):

release_collection.py
nhosoi: > > Sorry, I missed it... (I tested with python3.6, but there was no problem reported... (;_;) I wish it failed there...):
> 
> How did you test with python 3.6? When I test, I'm doing something like this:

I see... I just ran release_collection.py with python3.6 on my laptop.
```
$ python3.6
Python 3.6.15 (default, Jun 10 2022, 00:00:00) 
```
Maybe, that was not good enough... I'm going to run the tests on podman from now on... Thanks!

PR #175, opened 2022-08-03
changelog from commit hashes; changelogs from spec and BZ

richm:
- add support for creating changelogs for rpm releases
- use git commit subjects for changelog if no version

### DIFF
--- a/bz-manage.sh
+++ b/bz-manage.sh
@@ -22,6 +22,7 @@ STATUS=${STATUS:-POST}
 BZ_URL=${URL:-https://bugzilla.redhat.com/buglist.cgi}
 LIMIT=${LIMIT:-100}
 BASE_URL="${BZ_URL}?limit=${LIMIT}&component=${COMPONENT}&${ITR_FIELD}=${ITR}"
+show_url="https://bugzilla.redhat.com/show_bug.cgi?id="
 
 setitmdtm() {
   local field val bzlist baseurl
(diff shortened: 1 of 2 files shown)


richm:
This goes with https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/79

Pretty much each packaging method has a different way to generate the collection changelog :-(

* galaxy - use release_collection.py - use each role CHANGELOG.md, or the git commits for commit hashes
* Fedora rpm - generate from the %changelog
* RHEL rpm and Automation Hub - generate using `bz-manage.sh rpm_release $newversion` - then edit CHANGELOG.md

nhosoi:
I'm playing with bz-manage.sh in this pr.
If I don't specify `INCLUDE_CLONE=true`, it returns 3 items.
If I do, then it returns none.
```
$ ITR=8.7.0 STATUS=POST INCLUDE_CLONE=true RPM_CL=true ./bz-manage.sh get_cl
$
```
Is it the expected behavior?

richm:
> I'm playing with bz-manage.sh in this pr. If I don't specify `INCLUDE_CLONE=true`, it returns 3 items. If I do, then it returns none.
> 
> ```
> $ ITR=8.7.0 STATUS=POST INCLUDE_CLONE=true RPM_CL=true ./bz-manage.sh get_cl
> $
> ```
> 
> Is it the expected behavior?

I think `get_cl` isn't useful any more - prefer to use `rpm_release`.  Also `INCLUDE_CLONE` isn't really useful either - for rhel spec files, the changelog should contain only the BZ for the given release - so you cannot include the clone BZ in the spec %changelog - you already could not include the clone BZ in the git commit message.

nhosoi:
Thank you, @richm. Yes, `rpm_release` works nicely.
I noticed that it generates `cl-md` like this. Without clicking the links, it'd make the users wonder why there are two same bz's listed. I guess we have to include it in a good practice to add [9.1.0] and [8.7.0] to the bz summary...
```
- [logging - [RFE] Support startmsg.regex and endmsg.regex in the files inputs](https://bugzilla.redhat.com/show_bug.cgi?id=2112143)
- [logging - [RFE] Support startmsg.regex and endmsg.regex in the files inputs](https://bugzilla.redhat.com/show_bug.cgi?id=2112145)
```

richm:
> Thank you, @richm. Yes, `rpm_release` works nicely. I noticed that it generates `cl-md` like this. Without clicking the links, it'd make the users wonder why there are two same bz's listed. I guess we have to include it in a good practice to add [9.1.0] and [8.7.0] to the bz summary...
> 
> ```
> - [logging - [RFE] Support startmsg.regex and endmsg.regex in the files inputs](https://bugzilla.redhat.com/show_bug.cgi?id=2112143)
> - [logging - [RFE] Support startmsg.regex and endmsg.regex in the files inputs](https://bugzilla.redhat.com/show_bug.cgi?id=2112145)
> ```

It isn't supposed to do that - it is only supposed to look at BZ from only 1 ITR - so there should never be two of the same bugs from different releases with different BZ numbers.

> 
> Plus if you could clean up README - adding `rpm_release` and removing `get_cl` (?), I'd appreciate it.

I have split this PR into https://github.com/linux-system-roles/auto-maintenance/pull/177 and https://github.com/linux-system-roles/auto-maintenance/pull/176

PR #176, opened 2022-08-03
use git commit subjects for changelog if no version

richm:
If we are using the commit hash to build the collection, there will
not be a version and a corresponding changelog.  In that case, use
the git commit titles and list them as Bug Fixes.
* This will inevitably be wrong - there will be New Features
and Other Changes listed erroneously - but there is no way to
know
* When there is an actual release with a version and a changelog,
some of the changelog information will be duplicated - the new
release will list some of the old changelog items again.

I think this is a minor problem to have as long as releases using
commit hashes are rare, and the changelog isn't really used in
Galaxy anyway.

Add support for `--use-commit-hash-role ROLENAME`.  For example,
if you want to do a collection release using the tag for every
role *except* sshd and network, you can do this:
```
./release_collection.py --use-commit-hash-role sshd \
  --use-commit-hash-role network
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,6 +472,12 @@ will publish the collection, waiting until it is completed.
   tagged, use this flag. However, this means you will not be able to use
   automatic versioning, and you will need to use `--new-version`, or manually
   edit the `collection_release.yml` and `galaxy.yml` and use `--no-update`.
+* `--use-commit-hash-role` - list - By default, use whatever is the value
+  of `--use-commit-hash`.  There are some cases where you want to use the
+  tag for all roles *except* one or a few.  In that case, you can use
+  `--use-commit-hash-role role1 --use-commit-hash-role role2` to specify the
+  roles for which you want to use the commit hash, and use the tag for all
+  other roles.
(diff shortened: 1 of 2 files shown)


nhosoi (dismissed):
lgtm

richm:
> lgtm

I found a bug - if you are going from a commit hash to a tag release, we need to know how to extract the right changelog e.g. at this line
```
            elif cl.startswith("[{}]".format(cur_ref)):
                break
```
if `cur_ref` is a commit hash, this will cause problems

richm:
> > lgtm
> 
> I found a bug - if you are going from a commit hash to a tag release, we need to know how to extract the right changelog e.g. at this line
> 
> ```
>             elif cl.startswith("[{}]".format(cur_ref)):
>                 break
> ```
> 
> if `cur_ref` is a commit hash, this will cause problems

Fixed - new version will get the last tag before the commit hash and use that tag to extract the changelog info.

nhosoi (approved):
lgtm

PR #182, opened 2022-08-11
lsr-role2collection.py - improves collection mapping

nhosoi:
- bug fix: README needs to be copied to the doc dir after the extra-mapping..
- improve --extra-mapping to support mapping fedora.linux_system_roles to redhat.rhel_system_roles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,9 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                        [--src-path SRC_PATH] [--src-owner SRC_OWNER] [--role ROLE]
                        [--new-role NEW_ROLE] [--replace-dot REPLACE_DOT]
                        [--subrole-prefix SUBROLE_PREFIX] [--readme README]
-                       [--extra-mapping EXTRA_MAPPING]
+                       [--extra-mapping EXTRA_MAPPING] [--meta-runtime META_RUNTIME]
+                       [--extra-script EXTRA_SCRIPT]
+
 ```
 
 ### optional arguments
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -194,6 +195,13 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
                      in the README that you do not want to change, you may have to change the
                      README in another way, not using this script, by using sed with a custom
                      regex.
+--collection-mapping If it is set and given namespace.collection is not default, the default
+                     collection fedora.linux_system_roles is converted to
+                     NAMESPACE.COLLECTION.
```
richm, line 200: I thought we could already do this with `--extra-mapping`?

README.md
```diff
@@ -194,6 +195,13 @@ lsr_role2collection.py [-h] [--namespace NAMESPACE] [--collection COLLECTION]
...
+--meta-runtime META_RUNTIME
+                     This is the path to the collection meta/runtime.yml - the default is
+                     /home/nhosoi/linux-system-roles/auto-
+                     maintenance/lsr_role2collection/runtime.yml.
```
richm, line 204: This doesn't look right

README.md
nhosoi: No, `--extra-mapping` unfortunately does not cover this case for now...

Maybe, I'd better improve `parse_extra_mapping` to handle it, instead...

README.md
nhosoi: ??? I just copied the Usage since it was missing... If it's not a good idea, I'll drop it. :)

richm: I mean you have `/home/nhosoi` there - should it be `$HOME/linux-system-roles/auto-maintenance/lsr_role2collection/runtime.yml` instead?

README.md
richm: ok - if `--extra-mapping` doesn't work, then add `--collection-mapping`

nhosoi: Thanks, @richm. I've modified the parser a bit to support it.

README.md
nhosoi: Ahhh. Sorry, I was blind... :p Then, I'm adding them back with $HOME...

lsr_role2collection.py
```diff
@@ -1088,7 +1088,9 @@ def role2collection():
             "This is a comma delimited list of extra mappings to apply when "
             "converting the files - this replaces the given role name with "
             "collection format with the optional given namespace and collection "
-            "as sell as the given FQCN with other FQCN."
+            "as sell as the given FQCN with other FQCN. In addition, 'fedora."
```
richm, line 1091: as well as

nhosoi: *shame shame* Thanks!

richm (approved):

nhosoi:
Sorry, @richm. I found a bug in this patch. :( Debugging it now...

PR #187, opened 2022-08-30
skip vault-variables.yml - not sure how to configure for vault

richm:
Skip parsing the file vault-variables.yml - there won't be any
modules or plugins in this file, or jinja numeric operations -
the file should contain only vault encrypted variables.


nhosoi (approved):
lgtm

PR #188, opened 2022-08-31
allow scanning of fqcn include_role, etc.

richm:
Look for FQCN for some ansible built-ins such as
ansible.builtin.include_role
ansible.builtin.import_role
ansible.builtin.include_vars
so that we can treat them the same way as the non-FQCN versions.


nhosoi (approved):
lgtm

PR #199, opened 2022-10-26
add support for token, cancel tasks

richm:


richm:
Until jenkins is fixed, I recommend running `cancel_duplicate_tasks` periodically

richm:
to get a token - login to jenkins - then see https://www.jenkins.io/doc/book/system-administration/authenticating-scripted-clients/
```


The API token is available in your personal configuration page. Click your name on the top right corner on every page, then click "Configure" to see your API token. (The URL $root/me/configure is a good shortcut.) You can also change your API token from here.
```

PR #201, opened 2022-10-27
release_collection.py - Create execution-environment.yml in the collection

nhosoi:
Note: put execution-environment.yml in the collection top dir instead of the `meta` dir since the referred files, `requirements.yml`, `requirements.txt`, and `bindep.txt`, are in the collection top.


nhosoi:
Thank you for your review, @richm!

PR #202, opened 2022-10-27
Update update_vendored_collections.py to upload sources to Fedora

spetrosi:
This PR needs Fedora password added to BaseOS CI as env var.
Can be merged after this PR is merged: https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/84#


richm:
We don't need the vendored sources on Fedora because we aren't doing any vendoring on Fedora?

update_vendored_collections/update_vendored_collections.py
```diff
@@ -87,14 +87,26 @@ def spec_replace_sources(content, collection_tarballs):
     return content
 
 
+def spec_replace_bundled_provides(content, collection_tarballs):
+    print("Replacing bundled provides in the spec file")
+    for collection, tarball in collection_tarballs.items():
+        version = re.sub(".tar.gz", "", re.sub(".*-", "", tarball))
```
richm, line 93:
```suggestion
        version = re.sub("[.]tar.gz", "", re.sub(".*-", "", tarball))
```
Use `[.]` to ensure matching a literal `.` (you could also use `\`)

spetrosi: Thank you, done. Flake was complaining about such escaping, I needed to add `r` to the string like `r"\.tar\.gz"`.

pcahyna:
> We don't need the vendored sources on Fedora because we aren't doing any vendoring on Fedora?

Unfortunately, they are needed, because the SRPM generation step is not allowed to know that, because it is not allowed to have conditional Sources (for good reasons). All the tarballs must be included in the SRPM, and the lookaside cache is the only place where they can be obtained from.

Whether special care is needed to continuously update them is another question. I suppose that this helps with the maintenance of the RHEL package.

spetrosi:
> > We don't need the vendored sources on Fedora because we aren't doing any vendoring on Fedora?
> 
> Unfortunately, they are needed, because the SRPM generation step is not allowed to know that, because it is not allowed to have conditional Sources (for good reasons). All the tarballs must be included in the SRPM, and the lookaside cache is the only place where they can be obtained from.
> 
> Whether special care is needed to continuously update them is another question. I suppose that this helps with the maintenance of the RHEL package.

How soon do you think you will implement Packit? We can either merge this change now and the corresponding change in ansible-baseos-ci 2975 or prepare for your workflow now.

update_vendored_collections/update_vendored_collections.py
```diff
@@ -87,13 +87,28 @@ def spec_replace_sources(content, collection_tarballs):
...
+    print("Replacing bundled provides in the spec file")
+    for collection, tarball in collection_tarballs.items():
+        version = re.sub(r"\.tar\.gz", "", re.sub(".*-", "", tarball))
+        content = re.sub(
+            r"bundled\(ansible-collectionbundled\(ansible-collection\("
+            + collection
+            + r"\)\) = .*",
+            "bundled(ansible-collection(" + collection + ")) = " + version,
+            content,
+            flags=re.MULTILINE,
```
richm, line 100: Can you give me an example of the string you are trying to match?  I'm not sure what `bundled\(ansible-collectionbundled\(ansible-collection\(" + collection + r"\)\) = .*"` is supposed to match

update_vendored_collections/update_vendored_collections.py
```diff
@@ -104,6 +119,44 @@ def spec_add_changelog(content, collection_tarballs, lines):
...
 
 
+def configure_fkinit(credentials_file, repo):
+    print("Configure Fedora kerberos")
+    hsh = yaml.safe_load(open(credentials_file))
+    credentials = {}
+    for credential in hsh["credentials"]:
+        credentials.update(credential)
+    cmd = ["fkinit", "--user", credentials["username"]]
+    run_cmd(cmd, repo, credentials["password"])
```
richm, line 129: Is there no token based access?  It has to be kinit?

richm: How does packit do this?  Does it also use kinit?

update_vendored_collections/update_vendored_collections.py
spetrosi: `Provides: bundled(ansible-collection(ansible.posix)) = 1.4.0`

update_vendored_collections/update_vendored_collections.py
spetrosi: Yes it requires Fedora token, I am adding it to BaseOS CI Openshift. It will then do `sed -i "s/<password>/${FEDORA_PASSWORD}/" fedora_credentials.yml`

spetrosi: And this PR adds `update_vendored_collections/fedora_credentials.yml`. Openshift will have the token stored in ${FEDORA_PASSWORD}

pcahyna:
> > > We don't need the vendored sources on Fedora because we aren't doing any vendoring on Fedora?
> > 
> > 
> > Unfortunately, they are needed, because the SRPM generation step is not allowed to know that, because it is not allowed to have conditional Sources (for good reasons). All the tarballs must be included in the SRPM, and the lookaside cache is the only place where they can be obtained from.
> > Whether special care is needed to continuously update them is another question. I suppose that this helps with the maintenance of the RHEL package.
> 
> How soon do you think you will implement Packit? We can either merge this change now and the corresponding change in ansible-baseos-ci 2975 or prepare for your workflow now.

Not sure, but note that we can get rid of vendored collections in Fedora entirely and then we don't need the new workflow (nor this change) to keep them up to date. See https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/100 for an example how it would be done.

update_vendored_collections/update_vendored_collections.py
pcahyna: Packit uses a token when used as a CLI: https://packit.dev/docs/configuration/#values . I don't know about Packit as a service, but I suppose it is using a token as well, storing the Kerberos password somewhere sounds like a bad idea.

richm:
> Not sure, but note that we can get rid of vendored collections in Fedora entirely and then we don't need the new workflow (nor this change) to keep them up to date. See https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/100 for an example how it would be done.

Sounds good to me.  I think https://src.fedoraproject.org/rpms/linux-system-roles/pull-request/100 can be merged, then this #202 PR can be reworked.

pcahyna:
@richm I think the PR should be closed, as the task it is doing is not useful anymore. The branch should be kept, in case some of the code can be reused for internal automation.
Not sure what to do for internal repos - do you want to automate such updates? That is, do you want to update bundled collections in RHEL to their newest versions proactively?

richm:
> @richm I think the PR should be closed, as the task it is doing is not useful anymore. The branch should be kept, in case some of the code can be reused for internal automation.

OK.

> Not sure what to do for internal repos - do you want to automate such updates? That is, do you want to update bundled collections in RHEL to their newest versions proactively?

Yes, but it would have to be a purely internal process, and it cannot be completely automated because of the BZ and other gating checks.

PR #216, opened 2022-11-01
adding a new role - use ref: null; added ad_integration and podman

richm:
Added ad_integration and podman roles.

When you want to add a new role to the collection, add it in alphabetical order
to collection_release.yml with a `ref: null`.  The next time you use
release_collection.py to build the collection, it will update the ref with the
correct tag.  Also, the changelog will say
```
### New Features
...
- ROLENAME - New Role
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,15 @@ under the github organization, and the default is the `ROLENAME`, so only use
 this if you need to specify a different name for the role in the collection.
 The `sshd` role currently uses both `org` and `repo`.
 
+To add a new role - add it with a `ref: null` e.g.
+```yaml
+ad_integration:
+  ref: null
+```
+
+Then the next time you run `release_collection.py`, it will know that this is a
+new role and will update the versions accordingly.
+
 Example:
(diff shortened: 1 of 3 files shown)


pcahyna:
@richm how does it know what is the correct tag? Does it sort all tags that looks like version numbers and takes the highest one?

richm:
> @richm how does it know what is the correct tag?

There is an implicit assumption here that the only git tags are semantic version tags, and the tags have been applied in the correct order.  With this assumption, the code can use `git describe --tags --long --abbrev=40` to get the following tuple: latest tag, latest commit hash, and the number of commits applied after latest tag.  If number of commits is `0`, this means latest tag == latest commit.  https://github.com/linux-system-roles/auto-maintenance/pull/216

> Does it sort all tags that looks like version numbers and take the highest one?

No, because so far the assumptions above are correct.  However, if we ever start having arbitrary tags on roles, then we'll need to do something like: get the list of tags, then sort using `parse_version` for collation.

PR #228, opened 2022-12-22
Number the roles according to their order in RPM spec

pcahyna:
Also, sort collection_release.yml according to this number, rather than alphabetically. Having the file sorted by sourcenum makes it easier to add new roles - otherwise it is difficult to figure out what is the last number used.

### DIFF
--- a/release_collection.py
+++ b/release_collection.py
@@ -700,7 +700,7 @@ def update_collection(args, galaxy, coll_rel):
         update_galaxy_version(args, galaxy, versions_updated)
         if not args.no_update:
             with open(args.collection_release_yml.name, "w") as crf:
-                yaml.safe_dump(coll_rel, crf, sort_keys=True)
+                yaml.safe_dump(coll_rel, crf, sort_keys=False)
 
     # If to-be-appended changelogs are found, update COLLECTION_CHANGELOG.md
     # and copy it to the collection docs dir.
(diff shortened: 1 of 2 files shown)


release_collection.py
```diff
@@ -700,7 +700,7 @@ def update_collection(args, galaxy, coll_rel):
         update_galaxy_version(args, galaxy, versions_updated)
         if not args.no_update:
             with open(args.collection_release_yml.name, "w") as crf:
-                yaml.safe_dump(coll_rel, crf, sort_keys=True)
+                yaml.safe_dump(coll_rel, crf, sort_keys=False)
```
richm, line 703: Does `yaml.safe_dump` preserve the order of the keys in `coll_rel`?  If it does not, is there a way to use `OrderedDict` for `coll_ref`?
What is the default value of `sort_keys`?  Can you omit the argument if `False` is the default?

pcahyna: With `sort_keys=False`, it preserves the order of keys. Plain Python dicts preserve the insert order of keys, so OrderedDict is unnecessary.
`sort_keys` is unfortunately undocumented. The ruamel.yaml overview https://yaml.readthedocs.io/en/latest/overview.html says "block style and key ordering are kept, so you can diff the round-tripped source" which suggests that preserving key ordering is the default. But, my experiment revealed that if I omit the argument, the YAML dict gets sorted. So one can not omit it.

richm (approved):
lgtm - ready to merge?

PR #230, opened 2023-01-03
Spec file helpers, to be used by Packit

pcahyna:
- Script to generate RPM changelog entry, suitable for Packit's changelog-entry action.

- Script to generate Sources lines and %setup line in spec. Expects special # BEGIN ... # END ... pairs of markers in the spec file and replaces the code between them by code generated from collection_release.yml. Currently recognized markers are: 
  - AUTOGENERATED SOURCES - generates the rolenameN macros and %deftag directives for all the roles in collection_release.yml and the SourceN: tags
  - AUTOGENERATED SETUP - generates the %setup line with all the SourceN numbers.

- Script to extract collection version from galaxy.yml.


generate_spec_sources.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+
+def generate_sources(output):
+    k = list(rolesbysourcenum.keys())
+    k.sort()
+
+    for i in k:
+        r = rolesbysourcenum[i]
+        org = roles[rolesbysourcenum[i]].get("org")
```
richm, line 83:
```suggestion
        org = roles[r].get("org")
```
unless semantically you want to differentiate between role `r` and `rolesbysourcenum[i]`
also below

generate_spec_sources.py
```diff
@@ -0,0 +1,134 @@
...
+        ref = roles[rolesbysourcenum[i]]["ref"]
+        if org:
+            # If we support other hosting service than Github, the format
+            # of collection_release.yml will have to change a bit.
+            output.write(f"%global forgeorg{i} https://github.com/{org}\n")
+        if repo:
+            output.write(f"%global repo{i} {repo}\n")
+        output.write(
+            f"""%global rolename{i} {r}
+%deftag {i} {ref}
```
richm, line 94: This assumes `ref` from `collection_release.yml` is a tag which is a semantic version - what if it is a commit hash?

pcahyna: Indeed - I have thought that the file is maintained by release_collection.py, which keeps track of original role versions, and thus automatically ensures that the refs are tags. But now I see that there is `--use-commit-hash-(role)`, which breaks this rule, although it is not in use currently.
Since release_collection.py knows whether it is using a commit hash or a tag as a ref, I would prefer to change the format of `collection_release.yml` a bit and let release_collection.py record this information, instead of forgetting it on write and then having to guess it here.

pcahyna: Thank you for the explanation how it would commit hashes be used. I implemented the support for them.
That said, when I try to use a commit hash in collection_release.yml, it seems to break the  "Publish fedora.linux_system_roles collection to Ansible Galaxy" GitHub action. It aborts with this:
```
Exception: No new-version given and cannot auto-update version.  Please examine the changes to the collection, determine what the new semantic version will be, and re-run this script with the --new-version argument
  Error: Process completed with exit code 1.
```
https://github.com/pcahyna/auto-maintenance/actions/runs/3878083249/jobs/6613862726
The action is heavily modified by me, but I don't think that the error is due to this.
So, I suspect that this "rare event" of using a commit hash could break the automation if you ever have to resort to it.

richm (dismissed):

richm:
Looks like you need to rebase on top of the latest `master` branch - otherwise, lgtm - is this ready to be merged?

pcahyna:
yes, it is ready - should I squash the corrections into the first commit while at it?

richm:
> yes, it is ready - should I squash the corrections into the first commit while at it?

Yes, please

pcahyna:
rebased

richm (approved):

PR #231, opened 2023-01-10
Tag and release on GitHub when publishing new collection version

pcahyna:
The release artifact will not contain the actual roles, only collection_release.yml that specifies their version, build helpers, and common files like changelog and metadata (galaxy.yml).
The main point of creating the GitHub release is to trigger Packit automation that will create a downstream (Fedora) RPM  build.
It will be also nice from a reproducibility point of view - it will provide a clear track record of which commit was used to build a particular collection version.

Other changes:
- Collection is published after commit and push. If we publish the collection first and the push then fails, we lose the sources that the collection was generated from. It is better to let the process fail before the collection gets published.
- Corrected one syntax error in GitHub action command (should be `::error::`, not `::error `)
- Do not use `find` to locate the collection artifact - since we know what will be the name of the resulting collection artifact, use it directly.

### DIFF
--- a/.github/workflows/publish_collection.yml
+++ b/.github/workflows/publish_collection.yml
@@ -32,21 +32,32 @@ jobs:
         shell: bash
         run: |
           set -euxo pipefail
+          OLDVER=$(python3 ./get_collection_version.py)
+          echo ::debug::"Old collection version is $OLDVER"
           # Ensure there is no dest_path before running release_collection.py
           dest_path=/var/tmp/collection
           rm -rf "$dest_path"
           echo ::group::Build Collection
           python3 ./release_collection.py --debug --dest-path "$dest_path"
           echo ::endgroup::
+          NEWVER=$(python3 ./get_collection_version.py)
           # We are up to date - exit
           if git diff --quiet; then
             echo ::info No roles have new releases - no collection will be published
             exit 0
+          elif [ "$OLDVER" == "$NEWVER" ]; then
+            echo ::error::"Collection version $OLDVER has not changed"
+            exit 1
+          else
+            echo ::debug::"New collection version $NEWVER will be released"
           fi
           # A new collection has been build - find the tarball
           _tarballs=( $(find "$dest_path" -maxdepth 1 -type f -name "*.tar.gz") )
           if [ "${#_tarballs[@]}" -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_collection.yml
```diff
@@ -30,35 +30,70 @@ jobs:
...
           echo ::group::Push updates
           git config user.name systemroller
           git config user.email "systemroller@users.noreply.github.com"
           git commit -a -m "Collection version was updated"
           git push
           echo ::endgroup::
+          # Publish the collection. This step should
+          # be last since the previous steps must succeed
+          echo ::group::Publish Collection to Galaxy
+          ansible-galaxy collection publish -vv --token "${{ secrets.GALAXY_API_KEY }}" "${_tarballs[0]}"
```
richm, line 72:
```suggestion
          ansible-galaxy collection publish -vv --token "${{ secrets.GALAXY_API_KEY }}" "$_tarball"
```

.github/workflows/publish_collection.yml
```diff
@@ -30,35 +30,70 @@ jobs:
...
           echo ::endgroup::
+          NEWVER=$(python3 ./get_collection_version.py)
           # We are up to date - exit
           if git diff --quiet; then
             echo ::info No roles have new releases - no collection will be published
+            echo "updated=0" >> $GITHUB_OUTPUT
             exit 0
+          elif [ "$OLDVER" == "$NEWVER" ]; then
+            echo ::error::"Collection version $OLDVER has not changed"
+            exit 1
```
richm, line 52: Does this mean the pipeline will report an error in this case?

pcahyna: In fact, yes. This should not happen. release_collection.py should have updated the collection version, if it has not, I consider it a bug.
Do you have a scenario where not updating the version number is valid?

.github/workflows/publish_collection.yml
pcahyna: Ooops. I did my test without the ansible-galaxy stuff. Thanks.

.github/workflows/publish_collection.yml
richm: > In fact, yes. This should not happen. release_collection.py should have updated the collection version, if it has not, I consider it a bug. Do you have a scenario where not updating the version number is valid?

ok - just wanted to confirm this is a deliberate error

richm (approved):

PR #239, opened 2023-01-26
koji and bodhi stuff must be done in dist-git

richm:
The koji and bodhi stuff are done in dist-git - see
https://src.fedoraproject.org/rpms/linux-system-roles/blob/rawhide/f/.packit.yaml


pcahyna:
thank you @richm for finding and fixing this.

PR #242, opened 2023-02-01
md2html.sh add functionality to add TOC to HTML

spetrosi:


md2html.sh
```diff
@@ -29,9 +36,17 @@ for file in "$@"; do
...
   if [ "$md2html_tool" == kramdown ]; then
+    if [ "$add_toc" -ne 0 ]; then
+      sed -i '1s/^/* toc\n{:toc}\n/' "${file}"
+    fi
     # Set locale to UTF-8 because by default it is set to US-ASCII
-    LC_ALL=C.UTF-8 $md2html_tool --extension parser-gfm --input GFM \
-    --output html "${file}" > "${file%.md}.html"
+    LC_ALL=C.UTF-8 $md2html_tool --extension parser-gfm \
+      --input GFM \
+      "$toc_arg" \
```
richm, line 45:
```suggestion
      $toc_arg \
```
otherwise you will be passing an argument that is an empty string to md2html
if shellcheck doesn't like it, add a suppression

richm (approved):

richm:
looks like python black was updated and is not happy with some of our python scripts

PR #257, opened 2023-03-30
Add lsr_fingerprint.py

nhosoi:
`lsr_fingerprint` scans files in the `templates` dir under `./role["lsrrolename"]`, if the file contains a string `role["reponame"]:role["rolename"]`, replaces it with `system_role:role["lsrrolename"]`. Note: `role` is an entry of `roles` that is defined as an array of dictionaries in `lsr_fingerprint.py`.

For example, in metrics, `performancecopilot:ansible-pcp` is replaced with `system_role:metrics`; in sshd, `willshersystems:ansible-sshd` is with `system_role:sshd`.

### DIFF
--- a/lsr_fingerprint.py
+++ b/lsr_fingerprint.py
@@ -33,6 +33,7 @@ roles = [
 changedlist = []
 # dirs in the current dir
 dirs = [d for d in listdir(".") if isdir(d)]
+exit_code = 0
 for d in dirs:
     for role in roles:
         if role["lsrrolename"] == d:
--- a/lsr_fingerprint.py
+++ b/lsr_fingerprint.py
@@ -14,6 +14,7 @@ E.g., in metrics, "performancecopilot:ansible-pcp" is replaced with
 "system_role:metrics" in roles/bpftrace/templates/bpftrace.conf.j2.
 """
 
+import sys
 from os import listdir, walk
 from os.path import join, isfile, isdir
 
--- a/lsr_fingerprint.py
+++ b/lsr_fingerprint.py
@@ -54,11 +54,19 @@ for d in dirs:
                                     lines = ifp.readlines()
                                 count = 0
                                 newlines = ""
+                                changed = {}
                                 for line in lines:
                                     newline = line.replace(oldpair, newpair)
                                     if newline != line:
+                                        if len(changed) > 0:
+                                            raise Exception(
+                                                "Error: duplicate fingerprint '{0}' found in {1}:{2}".format(
+                                                    line.strip(),
+                                                    tmplpath,
+                                                    count,
+                                                )
+                                            )
                                         changed = {
-                                            "role": role,
                                             "path": tmplpath,
                                             "linenumber": count,
                                             "oldline": line.strip(),
@@ -69,10 +77,10 @@ for d in dirs:
                                     count += 1
                                 with open(tmplpath, "w") as ofp:
                                     ofp.writelines(newlines)
-
-if len(changedlist) > 0:
-    print("Done. Made the following changes:")
-    for changed in changedlist:
-        print("{0}:{1}".format(changed["path"], changed["linenumber"]))
-        print("  old line: {0}".format(changed["oldline"]))
-        print("  new line: {0}".format(changed["newline"]))
+            print(
+                "{0} role done. Made the following changes:".format(role["lsrrolename"])
+            )
+            for changed in changedlist:
+                print("{0}:{1}".format(changed["path"], changed["linenumber"]))
+                print("  old line: {0}".format(changed["oldline"]))
+                print("  new line: {0}".format(changed["newline"]))
--- a/lsr_fingerprint.py
+++ b/lsr_fingerprint.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2023, Red Hat, Inc.
+# SPDX-License-Identifier: MIT
+"""
+Modify fingerprint in spec file to be system_role:$rolename
+
+lsr_fingerprint.py -
+scans files in the templates dir under ./role["lsrrolename"],
+if the file contains a string role["reponame"]:role["rolename"],
+replaces it with system_role:role["lsrrolename"].
+
+E.g., in metrics, "performancecopilot:ansible-pcp" is replaced with
+"system_role:metrics" in roles/bpftrace/templates/bpftrace.conf.j2.
+"""
+
+from os import listdir, walk
+from os.path import join, isfile, isdir
+
+roles = [
+    {
+        "reponame": "willshersystems",
+        "rolename": "ansible-sshd",
+        "lsrrolename": "sshd",
+    },
+    {
+        "reponame": "performancecopilot",
+        "rolename": "ansible-pcp",
+        "lsrrolename": "metrics",
+    },
+]
+
+# dirs in the current dir
+dirs = [d for d in listdir(".") if isdir(d)]
+for d in dirs:
+    for role in roles:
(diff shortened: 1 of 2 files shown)


nhosoi:
Tested adding these lines at the top of the %build section of the spec file and verified the `performancecopilot:ansible-pcp` is replaced with `system_role:metrics`, and `willshersystems:ansible-sshd` is with` system_role:sshd`.
```
@@ -457,6 +459,9 @@ done
 sh md2html.sh $readmes
 %endif
 
+# Modify fingerprint in spec file to be system_role:$rolename
+python lsr_fingerprint.py
+
 mkdir .collections
```

lsr_fingerprint.py
```diff
@@ -0,0 +1,55 @@
...
+    },
+]
+
+# dirs in the current dir
+dirs = [d for d in listdir(".") if isdir(d)]
+for d in dirs:
+    for role in roles:
+        if role["lsrrolename"] == d:
+            for root, dirs, files in walk(d):
+                for dir in dirs:
```
spetrosi, line 39: Is it fine that `dirs` repeats in the main parent loop and in this child loop?

nhosoi: Hi @spetrosi!

The main loop (at line 35) is for selecting `metrics` and `sshd`, and the child loop is for `sshd/templates/` and
```
metrics/roles/bpftrace/templates/	metrics/roles/pcp/templates/
metrics/roles/elasticsearch/templates/	metrics/roles/redis/templates/
metrics/roles/grafana/templates/	metrics/roles/repository/templates/
metrics/roles/mssql/templates/		metrics/roles/spark/templates/
metrics/roles/pcp/meta/
```
in each role. As we have to support the vendored roles in `metrics`, the child loop is recursive.

Thanks!

richm, line 39:
```suggestion
                for dir in subdirs:
```

nhosoi: Ah, right. Thank you, @richm.

richm:
looking good - one more thing - please have the script print the changes made - something like this:
```
Done.  Made the following changes:
$lsrrolename/templates/$file:$linenumber
old line:.....
new line:.....
$lsrrolename/templates/$file:$linenumber
....
```
That way we can tell if the template had $repo:$role in some strange, unexpected location

richm:
also - each file should have 1 and only 1 place where this replacement is done - I think the script should be changed to raise an exception if more than 1 replacement is done - that is, it should first report all of the lines that had changes, then raise an exception, or exit with an error code, if the number of replacements is more than one.  That should help us catch odd unexpected cases.

lsr_fingerprint.py
```diff
@@ -44,12 +47,40 @@
...
-                            newlines = lines.replace(
-                                "{0}:{1}".format(role["reponame"], role["rolename"]),
-                                "system_role:{0}".format(role["lsrrolename"]),
-                            )
-                            if lines != newlines:
-                                with open(tmplpath, "w") as fp:
-                                    fp.write(newlines)
+                            with open(tmplpath) as ifp:
+                                lines = ifp.read()
+                            if lines.find(oldpair):
```
richm, line 52: `find` returns `-1` if the string was not found - https://docs.python.org/3/library/stdtypes.html#str.find

lsr_fingerprint.py
```diff
@@ -44,12 +47,40 @@
...
+                                with open(tmplpath) as ifp:
+                                    lines = ifp.readlines()
+                                count = 0
+                                newlines = ""
+                                changed = {}
+                                for line in lines:
+                                    newline = line.replace(oldpair, newpair)
+                                    if newline != line:
+                                        if len(changed) > 0:
+                                            raise Exception(
```
richm, line 62: This will exit the script without processing any other lines or files - so you won't know how many other lines or files have more than 1 `oldpair` in them

lsr_fingerprint.py
```diff
@@ -44,12 +47,40 @@
...
+                                            "path": tmplpath,
+                                            "linenumber": count,
+                                            "oldline": line.strip(),
+                                            "newline": newline.strip(),
+                                        }
+                                        changedlist.append(changed)
+                                    newlines = "{0}{1}".format(newlines, newline)
+                                    count += 1
+                                with open(tmplpath, "w") as ofp:
+                                    ofp.writelines(newlines)
```
richm, line 79: `writelines` is typically used with a `list` or other iterable - I guess it would work with a `str` type, but it's confusing that that argument to `writelines` is a `str`.  Could you do
```
56: newlines = []
...
76: newlines.append(newline)
...
ofp.writelines(newlines)
```

lsr_fingerprint.py
```diff
@@ -0,0 +1,92 @@
...
+                            tmplpath = join(dirpath, tmpl)
+                            with open(tmplpath) as ifp:
+                                lines = ifp.read()
+                            if oldpair in lines:
+                                with open(tmplpath) as ifp:
+                                    lines = ifp.readlines()
+                                count = 0
+                                newlines = []
+                                changed = {}
+                                for line in lines:
```
richm, line 58:
```suggestion
                                for line, lineno in enumerate(lines):
```
https://docs.python.org/3/library/functions.html#enumerate
then get rid of `count = 0` and `count += 1`

lsr_fingerprint.py
```diff
@@ -0,0 +1,92 @@
...
+                                    newline = line.replace(oldpair, newpair)
+                                    if newline != line:
+                                        if len(changed) > 0:
+                                            dup = True
+                                        else:
+                                            dup = False
+                                        changed = {
+                                            "path": tmplpath,
+                                            "linenumber": count,
```
richm, line 67:
```suggestion
                                            "linenumber": lineno,
```

lsr_fingerprint.py
nhosoi: Thank you, @richm! Updated.

lsr_fingerprint.py
```diff
@@ -0,0 +1,90 @@
...
+    {
+        "reponame": "performancecopilot",
+        "rolename": "ansible-pcp",
+        "lsrrolename": "metrics",
+    },
+]
+
+changedlist = []
+# dirs in the current dir
+dirs = [d for d in listdir(".") if isdir(d)]
```
richm, line 35:
```suggestion
dirs = [d for d in listdir(".") if isdir(d)]
exit_code = 0
```

lsr_fingerprint.py
```diff
@@ -0,0 +1,90 @@
...
+                        "DUPLICATED - " if changed["duplicate"] else "",
+                        changed["path"],
+                        changed["linenumber"],
+                    )
+                )
+                if changed["duplicate"]:
+                    print("  questionable line: {0}".format(changed["oldline"]))
+                else:
+                    print("  old line: {0}".format(changed["oldline"]))
+                    print("  new line: {0}".format(changed["newline"]))
```
richm, line 90:
```suggestion
                    print("  new line: {0}".format(changed["newline"]))
sys.exit(exit_code)
```

lsr_fingerprint.py
```diff
@@ -0,0 +1,90 @@
...
+
+lsr_fingerprint.py -
+scans files in the templates dir under ./role["lsrrolename"],
+if the file contains a string role["reponame"]:role["rolename"],
+replaces it with system_role:role["lsrrolename"].
+
+E.g., in metrics, "performancecopilot:ansible-pcp" is replaced with
+"system_role:metrics" in roles/bpftrace/templates/bpftrace.conf.j2.
+"""
+
```
richm, line 16:
```suggestion

import sys
```

richm (approved):

PR #265, opened 2023-04-26
build: change fingerprints in tests too

richm:
change the fingerprints in the test code.  Allow duplicates
in the test code.


richm:
This is to support https://github.com/performancecopilot/ansible-pcp/pull/51 and https://github.com/willshersystems/ansible-sshd/pull/232

spetrosi (approved):

PR #274, opened 2023-06-02
Collect PR descriptions instead of commmits into the changelog using API

spetrosi:
Developers write commits from a developer perspective and for developers, not for users. Sometimes there must be several commits in a single PR, and sometimes commits requre technical description. End users are not interested in technical details, they want to get information about how they benefit from code changes.
Hence, instead of collection commits into changelog, let's collect PR titile and description.

### DIFF
--- a/role-make-version-changelog.sh
+++ b/role-make-version-changelog.sh
@@ -7,43 +7,22 @@ set -euo pipefail
 
 AUTOSKIP="${AUTOSKIP:-true}"
 # Find the last tag in each role
-# Look at the git commits since that tag
-# Look at the actual changes since that tag
-# Figure out what to use for the new tag
-# Figure out what to put in the release notes for the release
+# Look at the merged PRs since that tag
+# Automatically figure out what version to use for the new tag
+# Automatically figure out what to put in the release notes for the release
 
-# By default, if there are commits that do not follow the conventional
+# By default, if there are PR titles that do not follow the conventional
 # commits format, the script will exit with an error.  You can set
-# ALLOW_BAD_COMMITS=true to continue to edit the CHANGELOG to deal with
-# the bad commits.
-ALLOW_BAD_COMMITS="${ALLOW_BAD_COMMITS:-false}"
+# ALLOW_BAD_PRS=true to continue to edit the CHANGELOG to deal with
+# the bad PR titles.
+ALLOW_BAD_PRS="${ALLOW_BAD_PRS:-false}"
 
-repo=${repo:-$(git remote get-url origin | awk -F'/' '{print $NF}')}
+repo_entries="$(gh repo view --json name,owner)"
+repo=${repo:-$(echo "$repo_entries" | jq -r .name)}
+owner=${owner:-$(echo "$repo_entries" | jq -r .owner.login)}
 
 # To be used in conjunction with local-repo-dev-sync.sh
 # This script is called from every role
-
-view_diffs() {
-    local tag="$1"
-    local action
-    action=""
-    while [ "$action" != "q" ]; do
-        action=""
-        for hsh in $(git log --pretty=tformat:%h --no-merges --reverse "${tag}.."); do
-            git show --stat "$hsh"
-            read -r -p 'View full diff (y)? Start over (s)? Quit viewing diffs (q)? Next commit (n)? (default: n) ' action
-            if [ "$action" = y ]; then
-                git show "$hsh"
-            elif [ "$action" = s ] || [ "$action" = q ]; then
-                break
-            fi
-            action=""
-        done
-        if [ -z "${action}" ]; then
-            break
-        fi
-    done
-}
 
 get_main_branch() {
     local br
@@ -59,13 +38,6 @@ get_main_branch() {
     fi
     echo UNKNOWN
     return 1
-}
-
-# Format commit as a gfm markdown bullet point
-# Keep a single new line at the end
-format_commit() {
-    git log --oneline --no-merges --reverse --pretty=format:"- %s%n%n%w(0,2,2)%b%n" -1 "$1" | \
-        awk 'NF > 0 {blank=0} NF == 0 {blank++} blank < 2'
 }
 
 if ! type -p npm > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


spetrosi:
@richm This requires some proper testing that I am about to do.
So far, please review the code.

For now, I am only changing the smallest part of the script to use PRs instead of commits. However, I think we can go further and automate this whole process, so that you push one button, and the script reads all repos and tags them. This is a next step, though.

role-make-version-changelog.sh
```diff
@@ -148,57 +133,47 @@ else
...
         echo ""
-        # get the commits since the tag
-        git log --oneline --no-merges --reverse "${latest_tag}"..
+        # get PR titles since the tag
+        tag_date=$(gh api repos/"$owner"/"$repo"/releases/latest -q .published_at)
+        commits=$(gh api repos/"$owner"/"$repo"/commits?since="$tag_date" -q '.[] | .sha ' | sort -r)
+        pr_titles_file=".pr_titles"
+        : > $pr_titles_file
+        for commit in $commits; do
+            pr_title=$(gh api repos/"$owner"/"$repo"/commits/"$commit"/pulls -q '(.[].title) + " #" + (.[].number | tostring)')
```
richm, line 144:
```suggestion
            pr_title=$(gh api repos/"$owner"/"$repo"/commits/"$commit"/pulls -q '.[] | (.title + " #" (.number | tostring)')
```

role-make-version-changelog.sh
```diff
@@ -148,57 +133,47 @@ else
...
-        echo Reword the commits that did not pass by marking them with "r"
-        echo Then perform a force push
-        exit 1
-    else
-        echo Success
-        echo ""
-    fi
+    echo Verifying if all PR titles comply with the conventional commits format
+    for commit in $commits; do
+        gh api repos/"$owner"/"$repo"/commits/"$commit"/pulls -q '.[].title' | npx commitlint --verbose
```
richm, line 162: should figure out how to use `$pr_titles_file` for this

richm:
This looks pretty good.   The queries are getting a bit complex.  Maybe this is ok - we'll have to see how this works for doing an update of many roles at once, but I think doing it this way should be ok.

We should look at using the graphql api instead of the REST api using `gh api graphql`.  I think we would be able to retrieve all of the pull requests that have been merged in a single query - https://docs.github.com/en/graphql/reference/objects#repository - `pullRequests` - still doesn't look like there is a simple query for "give me all of the pull requests that were merged after this timestamp" or "give me all of the pull requests that were merged after this tag/release" - but the nice thing about graphql is that you can retrieve all of the pull requests at once, then filter them out, instead of having to retrieve them one at a time.  Or, retrieve all of the commits at once, including the pulls.

role-make-version-changelog.sh
spetrosi: I didn't know that, thank you!

spetrosi:
I didn't know that GraphQL is a thing, GH docs somehow only mention REST in details, thank you for noting this. I found a way to list all PRs like you suggested. graphql API is faster as is, moreover, because with graphql the script don't need to run API calls many times in a loop it saves much time. I'll soon post another commits to use graphql

spetrosi:
GraphQL implementation is ready

role-make-version-changelog.sh
```diff
@@ -114,91 +87,122 @@ else
...
-        # network had a case where there where two tags for the
-        # latest - one with leading v and one without - git describe
-        # would return the one with the leading v - so had to strip
-        # it off
-        latest_tag="${latest_tag//v}"
-    fi
-    ;;
-esac
+# get latest release
+releases_latest=$(gh api repos/"$owner"/"$repo"/releases/latest -q '.tag_name,.published_at' 2> /dev/null || :)
```
richm, line 91:
```suggestion
read -r latest_tag release_timestamp <<< $(gh api repos/"$owner"/"$repo"/releases/latest -q '.tag_name + " " + .published_at' 2> /dev/null || :)
```
if the release exists:
```yaml
read -r latest_tag release_timestamp <<< $(gh api repos/linux-system-roles/network/releases/latest -q '.tag_name + " " + .published_at' 2>&1 || :)
echo $latest_tag
1.12.0
echo $release_timestamp
2023-05-30T19:34:49Z
```
if there is no release
```
read -r latest_tag release_timestamp <<< $(gh api repos/linux-system-roles/bootloader/releases/latest -q '.tag_name + " " + .published_at' 2>&1 || :)
echo $latest_tag
{"message":"Not
echo $release_timestamp
Found","documentation_url":"https://docs.github.com/rest/reference/repos#get-the-latest-release"}gh: Not Found (HTTP 404)
```
so you could still use a similar comparison for `if [[ "$release_timestamp" == *"Not Found"* ]]; then`

role-make-version-changelog.sh
```diff
@@ -114,91 +87,122 @@ else
...
-        # see the changes?
-        read -r -p 'View changes (n/y/s)? (default: n) ' view_changes
-        if [ "${view_changes:-n}" = y ]; then
-            view_diffs "${latest_tag}"
-        elif [ "${view_changes:-n}" = s ]; then
-            skip=true
+    latest_tag="$(echo "$releases_latest" | sed -n 1p)"
+    tag_range="closed:>$(echo "$releases_latest" | sed -n 2p)"
+fi
+# Note that first: 100 is a maximum, it would work only if there are <100 PRs since last tag
```
richm, line 106: yeah, I think this is fine.

role-make-version-changelog.sh
```diff
@@ -209,7 +213,7 @@ if [ "$skip" = false ]; then
             ver_minor=$((ver_minor+=1))
             ver_patch=0
         fi
-    elif grep -q '^[a-z0-9][a-z0-9]* feat.*' $rev_commits_file; then
+    elif grep -q '^[a-z0-9][a-z0-9]* feat.*' $pr_titles_file; then
```
richm, line 216: I guess this will be different - the format won't be `abcdef1 feat: something` since you don't have a commit hash anymore

richm:
Looks like if you use it like `role-manage-repos.sh /path/to/auto-maintenance/role-make-version-changelog.sh`, it does not clear the list of features and fixes from the previous roles - I'm doing a release for the storage role, and this list the list of features:
```
- feat: implement rhc_proxy.scheme (#106)
- feat: add AlmaLinux to RHEL compat distro list (#618)
- feat: Add support for setting stripe size for LVM RAID (#357)
```
the first is from rhc, the second from network - same for the list of fixes and other changes

role-make-version-changelog.sh
spetrosi: I'd need too look for `message":"Not` or `Found",` in `$latest_tag` or `$release_timestamp`, isn't it uglier than the initial solution? With `-q '.tag_name,.published_at'`, it prints

role-make-version-changelog.sh
spetrosi: Me too, 100 PRs is a lot

role-make-version-changelog.sh
richm: ok - then leave it - it's fine

spetrosi:
What is role-manage-repos.sh? I do not know how this can happen because the files $new_features_file, $bug_fixes_file, $other_changes_file are removed with `rm -f "$new_features_file" "$bug_fixes_file" "$other_changes_file"` prior to using them, and also reused at the end of the script.

richm:
> What is role-manage-repos.sh?

https://github.com/linux-system-roles/auto-maintenance/blob/main/manage-role-repos.sh
https://github.com/linux-system-roles/auto-maintenance#manage-role-repossh

I use it when I need to update several roles at the same time.

> I do not know how this can happen because the files $new_features_file, $bug_fixes_file, $other_changes_file are removed with `rm -f "$new_features_file" "$bug_fixes_file" "$other_changes_file"` prior to using them, and also reused at the end of the script.

Yeah, I'm not sure what's going on.  I need to do some investigating.

spetrosi:
That happened because `declare -a pr_descriptions` was not emptying an array as I was mistakenly thinking. I fixed it now by declaring an empty array with `pr_descriptions=()` this should be fine.

richm:
lgtm - rebase and we can merge

PR #280, opened 2023-06-29
skip files directory to avoid yml file errors

richm:
The podman role has test yml files that should not be processed
by lsr_role2collection or report-modules-plugins.  The lsr r2c
should skip processing any `files/` or `templates/` directories.
The `report-modules-plugins.py` should skip checking `files/`
files for plugins.

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -42,6 +42,7 @@ ALL_ROLE_DIRS = [
     "defaults",
     "examples",
     "files",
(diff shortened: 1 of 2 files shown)


richm:
see https://github.com/linux-system-roles/podman/actions/runs/5407050225/jobs/9824604133?pr=78
and https://github.com/linux-system-roles/podman/actions/runs/5407050230/jobs/9824604196?pr=78

PR #281, opened 2023-06-30
Add update_doc_text functionality to bz-manage.sh

spetrosi:
Now that all PRs in system roles follow the conventional commits format (see https://linux-system-roles.github.io/contribute.html#pr-description-format) we can automate updating doc texts in Bugzilla.

This PR expands the bz-manage.sh script with `update_doc_text`. To be run like `ITR=9.3.0 sh ./bz-manage.sh update_doc_text`.

When run, the script searches all PRs from the ITM and other filters provided and:
1.  If a BZ has required_doc_text flag set to `+` or `-`, the BZ is skipped. + indicates that the doc text has already been verified by a tech writer. - indicates that no release note is required. 
2. If a BZ has required_doc_text unset and GH PR set:
  Automatically add doc text from PR description
3. If a BZ has required_doc_text? set and GitHub PR set:
  If the existing doc_text is empty, automatically add doc text from PR description.
  Else, print the current doc text and doc type, print the GH PR info, and ask user if they want to update the doc text.
4. If a BZ has required_doc_text not set to + or - and GitHub PR unset:
  Print the current doc text and type, and ask user if they want to post a comment asking to attach a GH PR or enter the doc text.

### DIFF
--- a/bz-manage.sh
+++ b/bz-manage.sh
@@ -8,7 +8,6 @@ fi
 get_bzs() {
   # shellcheck disable=SC2086
   # jq arguments $3 are unquoted to allow full expansion
-  # echo "bugzilla query --from-url \"$1\" --json | jq ${3:-} \"${2:-.bugs[].id}\""
   bugzilla query --from-url "$1" --json | jq ${3:-} "${2:-.bugs[].id}"
 }
 
@@ -443,18 +442,22 @@ bz_mod_doc_text() {
       echo ""
       echo "Doc Type and Text suggested from the GH PR:"
       print_doc_text "$gh_pr_type" "$gh_pr_desc"
-      read -r -p 'Do you want to update the BZ with the suggested Doc Text and Type (y/n)? (default: n) ' update_bz </dev/tty
-      if [ "${update_bz:-n}" = n ]; then
-        return
-      elif [ "${update_bz:-n}" = y ]; then
-        comment="Updating Doc Type and Text with info from the attached PR $gh_pr"
-        echo "BZ#$bz: $comment"
-        bugzilla modify --private \
-          --field cf_doc_type="$gh_pr_type" \
-          --field cf_release_notes="$gh_pr_desc" \
-          --comment "$comment" \
-          "$bz"
-      fi
+      while read -r -p 'Do you want to update the BZ with the suggested Doc Text and Type (y/n)? (default: n) ' update_bz </dev/tty; do
+        if [ "${update_bz:-n}" = n ]; then
+          return
+        elif [ "${update_bz:-n}" = y ]; then
+          comment="Updating Doc Type and Text with info from the attached PR $gh_pr"
+          echo "BZ#$bz: $comment"
+          bugzilla modify --private \
+            --field cf_doc_type="$gh_pr_type" \
+            --field cf_release_notes="$gh_pr_desc" \
+            --comment "$comment" \
+            "$bz"
+          return
+        else
+          echo "$update_bz is not y or n"
+        fi
+      done
     else
       echo "BZ#$bz already has Doc Text that equals GitHub PR description"
     fi
@@ -462,9 +465,8 @@ bz_mod_doc_text() {
 }
 
 process_count_prs() {
-  bz=$1
-  func=$2
-  args=$3
+  bz="$1"; shift
+  func="$1"; shift
   echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
   jq_count_prs="select(.id == $bz)\
 | [.external_bugs[]\
@@ -472,8 +474,9 @@ process_count_prs() {
 | length"
   gh_pr_count=$(jq -r "$jq_count_prs" "$query_file")
   if [ "$gh_pr_count" -eq 1 ]; then
-    # shellcheck disable=SC2086
-    $func $args
+    # eval to re-evaluate arguments for the $func function
+    # shellcheck disable=SC2294
+    eval "$func" "$@"
   elif [ "$gh_pr_count" -gt 1 ]; then
     echo "BZ#$bz has $gh_pr_count GH PRs attached, you may choose info from what PR to use"
     for ((i=0;i<gh_pr_count;i++)); do
@@ -484,16 +487,21 @@ process_count_prs() {
       echo "($i) Doc Type and Text suggested from GH PR $gh_pr:"
       print_doc_text "$gh_pr_type" "$gh_pr_desc"
     done
-    read -r -p "What PR do you want to use for Doc Text and Type? (type a number from 0 to $((gh_pr_count-1)), or s to skip)? (default: s) " update_gh </dev/tty
-    if [ "${update_gh:-s}" = s ]; then
-      return
-    else
-      jq_gh_pr="\"https://github.com/\" + ([.external_bugs[]\
-| select((.type.type == \"GitHub\") and (.ext_bz_bug_id | contains(\"pull\")))][$update_gh] .ext_bz_bug_id | tostring)"
-      gh_pr=$(jq -r "$jq_gh_pr" "$query_file")
-      # shellcheck disable=SC2086
-      $func $args
-    fi
+    while read -r -p "What PR do you want to use for Doc Text and Type? (type a number from 0 to $((gh_pr_count-1)), or s to skip)? (default: s) " update_gh </dev/tty; do
+      if [ "${update_gh:-s}" = s ]; then
+        return
+      elif [ "${update_gh:-s}" -ge 0 ] && [ "${update_gh:-s}" -le "$((gh_pr_count-1))" ]; then
+        jq_gh_pr="\"https://github.com/\" + ([.external_bugs[]\
+  | select((.type.type == \"GitHub\") and (.ext_bz_bug_id | contains(\"pull\")))][$update_gh] .ext_bz_bug_id | tostring)"
+        gh_pr=$(jq -r "$jq_gh_pr" "$query_file")
+        # eval to re-evaluate arguments for the $func function
+        # shellcheck disable=SC2294
+        eval "$func" "$@"
+        return
+      else
+        echo "$update_gh is not a valid integer"
+      fi
+    done
   fi
 }
 
@@ -519,8 +527,10 @@ update_doc_text() {
 [(.external_bugs[] | select(.type.type == \"GitHub\") | select(.ext_bz_bug_id | contains(\"pull\")) | .ext_bz_bug_id | tostring)][0]\
 + \"|\" + .cf_doc_type"
   jq -r "$jq" "$query_file" | while IFS=\| read -r bz gh_pr doc_type; do
-    process_count_prs "$bz" bz_mod_doc_text "$gh_pr $bz $query_file"
-    done
+    # With the last argument, provide arguments to use for the function
+    # shellcheck disable=SC2016
+    process_count_prs "$bz" bz_mod_doc_text '$gh_pr $bz $query_file'
+  done
 
   # II. If a BZ has required_doc_text unset and GH PR set:
   # Add doc text from PR description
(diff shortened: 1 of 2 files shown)


bz-manage.sh
```diff
@@ -381,4 +382,137 @@ get_bz_id_summary() {
...
+    gh_pr_title=$(echo "$gh_pr_info" | sed -n 1p)
+    gh_pr_desc=$(echo "$gh_pr_info" | sed 1d)
+    if [[ "$gh_pr_title" =~ ^feat.* ]]; then
+      gh_pr_type="Enhancement"
+    elif [[ "$gh_pr_title" =~ ^fix.* ]]; then
+      gh_pr_type="Bug Fix"
+    fi
+    comment="Updating Doc Type and Text with info from the attached PR $gh_pr"
+    echo "BZ#$bz: $comment"
+    echo bugzilla modify --private \
```
spetrosi, line 410: How can I test this? The best thing I came up with is to open a BZ for my ansible-collection-microsoft-sql so that the minimal number of people receive notifications, is there a better way?
When writing this comment I came up with another thing - run bugzilla with `--minor-update` to not send updates on some old existing BZ. But still, there may be better options.

bz-manage.sh
```diff
@@ -381,4 +382,137 @@ get_bz_id_summary() {
...
+update_doc_text() {
+  local queryurl jq bz bz_assignee doc_type doc_text gh_pr gh_pr_info gh_pr_title gh_pr_desc gh_pr_type update_bz
+  query_file=.query_file.json
+
+  # If a BZ has required_doc_text unset and GH PR set:
+  # Automatically add doc text from PR description
+  queryurl="${BASE_URL}"
+  queryurl+="&f1=flagtypes.name&f2=external_bugzilla.url&f3=ext_bz_bug_map.ext_bz_bug_id"
+  # substring doesn't work with ext_bz_bug_map.ext_bz_bug_id, using anywordssubstr
+  queryurl+="&o1=notsubstring&o2=substring&o3=anywordssubstr"
```
spetrosi, line 394: Bugzilla does not search for links properly for some reason, I found a workaround with using anywordssubstr filter instead of substring. Apparently, this works. And for `not anywordssubstr` I set also n=1 in some below search.

bz-manage.sh
```diff
@@ -381,4 +382,137 @@ get_bz_id_summary() {
...
 }
 
+update_doc_text() {
+  local queryurl jq bz bz_assignee doc_type doc_text gh_pr gh_pr_info gh_pr_title gh_pr_desc gh_pr_type update_bz
+  query_file=.query_file.json
+
+  # If a BZ has required_doc_text unset and GH PR set:
+  # Automatically add doc text from PR description
+  queryurl="${BASE_URL}"
+  queryurl+="&f1=flagtypes.name&f2=external_bugzilla.url&f3=ext_bz_bug_map.ext_bz_bug_id"
```
spetrosi, line 392: Is there a better way to write a long string in multiple lines than doing += to the previous string?

bz-manage.sh
```diff
@@ -381,4 +382,137 @@ get_bz_id_summary() {
...
+    if [[ "$gh_pr_title" =~ ^feat.* ]]; then
+      gh_pr_type="Enhancement"
+    elif [[ "$gh_pr_title" =~ ^fix.* ]]; then
+      gh_pr_type="Bug Fix"
+    fi
+    comment="Updating Doc Type and Text with info from the attached PR $gh_pr"
+    echo "BZ#$bz: $comment"
+    echo bugzilla modify --private \
+      --field cf_doc_type="$gh_pr_type" \
+      --field cf_release_notes="$gh_pr_desc" \
```
spetrosi, line 412: The $gh_pr_desc is always multiple lines of text. How do I make it work here? I'd think I can use env variable, but bugzilla cli doesn't allow for it.

spetrosi (commented):
Hi @richm, WDYT? I tried to automate the most but ensure that automation don't screw us. Because if requires_doc_text is set to `?` and doc text exists we do not want to force overwrite, there can be some useful text there already.
I wrote some questions for you in the inline comments.
Thank you!

bz-manage.sh
richm:
```suggestion
  queryurl="${BASE_URL}&f1=flagtypes.name&f2=external_bugzilla.url&f3=ext_bz_bug_map.ext_bz_bug_id"
```
?

bz-manage.sh
richm: ok - works for me

bz-manage.sh
richm: I think you should open a BZ for testing purposes - just clearly label the summary like "ignore - for automation testing purposes only" or something like that - for PRs, you can use the ci-testing repo - and yes, the `--minor-update` flag might also be useful

bz-manage.sh
richm: You should be able to pass a multiline string in a double quoted variable

bz-manage.sh
richm: Maybe you mean - how to use line wrapping in bash?
```bash
queryurl="${BASE_URL}\
&f1=flagtypes.name&f2=external_bugzilla.url&f3=ext_bz_bug_map.ext_bz_bug_id\
&o1=notsubstring&o2=substring&o3=anywordssubstr"
```
then queryurl will be a very long single line string

spetrosi: And if the line is very long and I want to "wrap" it?

richm:
```bash
# foo="this is a\
very long string that is\
wrapped in a couple\
of places"
# echo "$foo"
this is avery long string that iswrapped in a coupleof places
```

bz-manage.sh
```diff
@@ -381,4 +382,203 @@ get_bz_id_summary() {
...
+          "$bz"
+      fi
+    else
+      echo "BZ#$bz already has Doc Text that equals GitHub PR description"
+    fi
+  fi
+}
+
+process_count_prs() {
+  bz=$1
```
richm, line 465:
```suggestion
  bz="$1"; shift
```

bz-manage.sh
```diff
@@ -381,4 +382,203 @@ get_bz_id_summary() {
...
+  func=$2
```
richm, line 466:
```suggestion
  func="$1"; shift
```

bz-manage.sh
```diff
@@ -381,4 +382,203 @@ get_bz_id_summary() {
...
+  args=$3
```
richm, line 467: using `shift` above means the remaining args are now in `"$@"` - now you don't need to use `args` - you can pass them as "$@" below

bz-manage.sh
```diff
@@ -381,4 +382,203 @@ get_bz_id_summary() {
...
+  echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+  jq_count_prs="select(.id == $bz)\
+| [.external_bugs[]\
+| select((.type.type == \"GitHub\") and (.ext_bz_bug_id | contains(\"pull\")))]\
+| length"
+  gh_pr_count=$(jq -r "$jq_count_prs" "$query_file")
+  if [ "$gh_pr_count" -eq 1 ]; then
+    # shellcheck disable=SC2086
+    $func $args
```
richm, line 476:
```suggestion
    "$func" "$@"
```

bz-manage.sh
```diff
@@ -381,4 +382,203 @@ get_bz_id_summary() {
...
+    done
+    read -r -p "What PR do you want to use for Doc Text and Type? (type a number from 0 to $((gh_pr_count-1)), or s to skip)? (default: s) " update_gh </dev/tty
+    if [ "${update_gh:-s}" = s ]; then
+      return
+    else
+      jq_gh_pr="\"https://github.com/\" + ([.external_bugs[]\
+| select((.type.type == \"GitHub\") and (.ext_bz_bug_id | contains(\"pull\")))][$update_gh] .ext_bz_bug_id | tostring)"
+      gh_pr=$(jq -r "$jq_gh_pr" "$query_file")
+      # shellcheck disable=SC2086
+      $func $args
```
richm, line 495:
```suggestion
      "$func" "$@"
```

bz-manage.sh
spetrosi: I was doing it incorrectly, now using eval to re-evaluate the variables

bz-manage.sh
spetrosi: I've applied this, but can you please elaborate on why this is better? Isn't it clearer to explicitly set `args=$3`. iiuc, the remaining args are in $3 anyway.

richm: > I've applied this, but can you please elaborate on why this is better? Isn't it clearer to explicitly set `args=$3`. iiuc, the remaining args are in $3 anyway.

It depends on how you want to use the args.  Consider this case:
```bash
foo() {
  ii=1
  while [ -n "${1:-}" ]; do
    echo "$ii" "$1"
    shift
    ii=$(expr "$ii" + 1)
  done
}
bar() {
  ii=1
  args="$1"
  for arg in $args; do
    echo "$ii" "$arg"
    ii=$(expr $ii + 1)
  done
}
foo "arg with spaces" "arg*" "another"
bar "arg\ with\ spaces arg\* another"
```
if you use the `foo` method, you preserve spaces and other shell metacharacters in the arguments, and you avoid having the `*` expand and possibly match files in the current directory.

bz-manage.sh
```diff
@@ -381,4 +381,220 @@ get_bz_id_summary() {
...
+  echo "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~"
+  jq_count_prs="select(.id == $bz)\
+| [.external_bugs[]\
+| select((.type.type == \"GitHub\") and (.ext_bz_bug_id | contains(\"pull\")))]\
+| length"
+  gh_pr_count=$(jq -r "$jq_count_prs" "$query_file")
+  if [ "$gh_pr_count" -eq 1 ]; then
+    # eval to re-evaluate arguments for the $func function
+    # shellcheck disable=SC2294
+    eval "$func" "$@"
```
richm, line 479: not sure why you need to use `eval` here - what does `eval` do vs. just doing `"$func" "$@"`?

spetrosi: Running with debug with eval:
```
+ eval bz_mod_doc_text '$gh_pr $bz $query_file'
++ bz_mod_doc_text https://github.com/linux-system-roles/auto-maintenance/pull/281 2219317 .query_file.json
```
Without:
```
+ bz_mod_doc_text '$gh_pr $bz $query_file'
+ gh_pr='$gh_pr $bz $query_file'
bz-manage.sh: line 425: $2: unbound variable
```

spetrosi: I tried to explain it in a comment above somehow, let me elaborate more here.
To `process_count_prs` with the last argument, I provide a list of arguments that I want to run with the $func function.
So I am running `process_count_prs "$bz" bz_mod_doc_text '$gh_pr $bz $query_file'` and then run `eval "$func" "$@"`.
If I use it like `process_count_prs "$bz" bz_mod_doc_text "$gh_pr $bz $query_file"` - pass last arg in double quotes, then `$gh_pr $bz $query_file` are populated with the initial values. But `process_count_prs` must re-evaluate `$gh_pr` value based on what PR serial number user provides.

This is needed because bz_add_doc_text takes 2 arguments - `$gh_pr $bz`, and `bz_mod_doc_text` takes 3 arguments - `$gh_pr $bz $query_file`.

As a workaround, in `process_count_prs`, I can also run func simply as `"$func" "$gh_pr" "$bz" "$query_file"`
It would naturally work for `bz_mod_doc_text` and won't break anything for `bz_add_doc_text` too, but would be hard to troubleshoot is something gets changed.

richm: I would prefer to find a solution that doesn't use `eval` but it may be unavoidable in this case.

What confuses me is that in https://github.com/linux-system-roles/auto-maintenance/pull/281/files#diff-e05dfb61babe0363643fe39c9ccdebea9f2e2fc34ad9c95b6fdf69b3eb748f60R529 you grab `gh_pr`, but you don't necessarily use it - in the case where `"$gh_pr_count" -gt 1` you may have multiple different values of `gh_pr`, and you want to use the value which is set in the loop at 496.

So I think this is ok

richm (approved):

PR #290, opened 2023-07-27
md2html: Add ability to select tool, always create toc, improve pandoc

spetrosi:
- You can now provide set md2html_tool to kramdown or pandoc to use a specific tool
- Toc is always created both on pandoc and kramdown
- Pandoc converts directly to html using a template


md2html.sh
```diff
@@ -17,44 +15,33 @@ while getopts "htl" opt; do
...
+    sed -i '1,2d' "${file}"
   elif [ "$md2html_tool" == pandoc ]; then
-    $md2html_tool -f gfm "${file}" -t asciidoc -o "${file%.md}.tmp.adoc"
-    touch -r "${file}" "${file%.md}.tmp.adoc"
-    TZ=UTC asciidoc -o "${file%.md}.html" -a footer-style=none -a toc2 -a source-highlighter=highlight "${file%.md}.tmp.adoc"
-    tr -d '\r' < "${file%.md}.html" > "${file%.md}.tmp.adoc"
-    mv "${file%.md}.tmp.adoc" "${file%.md}.html"
+    $md2html_tool --from gfm "${file}" --to html --output "${file%.md}.html" \
+      --toc --shift-heading-level-by=-1 \
+      --template https://raw.githubusercontent.com/tajmone/pandoc-goodies/master/templates/html5/github/GitHub.html5
```
richm, line 44: I guess if we ever want to use this downstream or in rpm build, we'll need to package this file

richm:
Can you provide an example of what this looks like?

md2html.sh
spetrosi: Oh yeah, I need to make this conditional so that we do not use this in RPM build for now. My plan was to get rid of it from the rpm build completely because there is not pandoc on RHEL 9+.

spetrosi:
Archive with mssql README previews for kramdown, pandoc-old, and pandoc-new that this PR introduces. [READMEs.zip](https://github.com/linux-system-roles/auto-maintenance/files/12183777/READMEs.zip)

spetrosi:
Last time I built kramdown without toc, here is with toc [READMEs.zip](https://github.com/linux-system-roles/auto-maintenance/files/12183805/READMEs.zip)

richm:
hmm - I don't like the pandoc-new - it only uses the middle third of the screen - very wide margins - I like the old style where there are no side margins

spetrosi:
> hmm - I don't like the pandoc-new - it only uses the middle third of the screen - very wide margins - I like the old style where there are no side margins

This can be fixed by removing `-width:980px;` from the template. I guess it'd be better to copy the template to auto-maintenance.

richm (approved):

spetrosi:
md2html will soon be obsolete after I remove it from spec. Instead, tox env should be used for convention, see https://github.com/linux-system-roles/tox-lsr/pull/133

PR #291, opened 2023-07-27
As part of creating new release, get .README.html from docs branch

spetrosi:
Keeping .README.html in each role's root dir would help to have a consistent .README.html in RPMs.
Keeping .README.html as a hidden file because it becomes outdated when new changes are made to README.md, and new release hasn't been created yet. The up-to-date README.html is stored in docs branch and is published to GitHub pages of each role as per https://github.com/linux-system-roles/.github/pull/41. 

See RHELPLAN-158622

### DIFF
--- a/role-make-version-changelog.sh
+++ b/role-make-version-changelog.sh
@@ -306,12 +306,10 @@ You have three options:
             cat "$rel_notes_file" CHANGELOG.md > .tmp-changelog
         fi
         mv .tmp-changelog CHANGELOG.md
-        # Expecting that this script is run from the cloned auto-maintenance repo
-        SCRIPT_DIR="$(dirname -- "$0")"
-        md2html_tool=pandoc sh "$SCRIPT_DIR"/md2html.sh README.md
-        git add CHANGELOG.md README.html
+        git show origin/docs:latest/README.html > .README.html
+        git add CHANGELOG.md .README.html
         { echo "docs(changelog): version $new_tag [citest skip]"; echo "";
-          echo "Update changelog and README.html for version $new_tag"; } > .gitcommitmsg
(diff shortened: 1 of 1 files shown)


role-make-version-changelog.sh
```diff
@@ -306,9 +306,12 @@ You have three options:
...
         fi
         mv .tmp-changelog CHANGELOG.md
-        git add CHANGELOG.md
+        # Expecting that this script is run from the cloned auto-maintenance repo
+        SCRIPT_DIR="$(dirname -- "$0")"
+        md2html_tool=pandoc sh "$SCRIPT_DIR"/md2html.sh README.md
+        git add CHANGELOG.md README.html
         { echo "docs(changelog): version $new_tag [citest skip]"; echo "";
-          echo "Create changelog update and release for version $new_tag"; } > .gitcommitmsg
+          echo "Update changelog and README.html for version $new_tag"; } > .gitcommitmsg
```
richm, line 314: I would also like to have a github action which creates README.html when README.md changes.
Note that the above might not change README.html, so the commit msg might be slightly misleading.

role-make-version-changelog.sh
```diff
@@ -306,9 +306,10 @@ You have three options:
             cat "$rel_notes_file" CHANGELOG.md > .tmp-changelog
         fi
         mv .tmp-changelog CHANGELOG.md
-        git add CHANGELOG.md
+        git show docs:latest/README.html > .README.html
```
richm, line 309: you can also use `git checkout $branchname -- $filename` to check the file out of the specified branch into the current branch, which will also do a `git add $filename`

spetrosi: In that case, I'll need to rename `README.html` > `.README.html` after checkouting it, and here it's one line.

richm (dismissed):

PR #292, opened 2023-07-27
Do not open editor when preparing CHANGELOG.md, it builds consistently

spetrosi:


richm:
I would still like to have the option to edit the changelog - for example, the network role does not want the full commit message in the changelog, only the PR title/summary line.

richm:
> Alrighty no worries

Well, if you want, you can add some sort of option/environment variable - if set to `true`, skip editing the changelog - default is `false` - use editor

PR #295, opened 2023-08-25
report-modules-plugins add local plugins from tests/library

spetrosi:
Specifically, the certificate role stores a module that is used for tests only in tests/library

### DIFF
--- a/report-modules-plugins.py
+++ b/report-modules-plugins.py
@@ -1179,10 +1179,10 @@ class SearchCtx(object):
 def usage():
     return f"""
     You must create an environment (e.g. python venv or VM) which uses
-    ansible 4.x and jinja2.7 e.g.
+    ansible 6.x, and jinja2.7 with python_version <= "3.7", e.g.
       python -mvenv .venv
       . .venv/bin/activate
-      pip install 'ansible==4.*' 'jinja2==2.7.*'
+      pip install 'ansible==6.*' 'jinja2==2.7.*; python_version <= "3.7"'
     Then run this script using that environment.
     Each argument is the name of a directory containing a role, a collection,
     a 'roles' directory containing multiple roles, or an 'ansible_collections'
--- a/report-modules-plugins.py
+++ b/report-modules-plugins.py
@@ -602,6 +602,7 @@ def process_role(role_path, is_real_role, ctx):
             ctx.found_role_name = get_role_name(role_path)
         ctx.enter_role(ctx.found_role_name, role_path)
         ctx.add_local_plugins(role_path, "library")
+        ctx.add_local_plugins(role_path, "tests/library")
         ctx.add_local_plugins(role_path, "filter_plugins")
     # process defaults to get role public api variables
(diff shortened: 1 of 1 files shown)


rjeffman:
Can the list of local plugins be defined per role?

richm:
> Can the list of local plugins be defined per role?

not sure what you mean - the script dynamically determines the list of plugins for each role

rjeffman:
> > Can the list of local plugins be defined per role?
> 
> not sure what you mean - the script dynamically determines the list of plugins for each role

This patch is adding another directory to search for plugins to a context, maybe this list of directories could be defined per role.

It is just an idea (maybe a bad one), thatk would help to fix similar issues in the future on any role by applying changes to the role configuration, not the testing tool.

richm:
> > > Can the list of local plugins be defined per role?
> > 
> > 
> > not sure what you mean - the script dynamically determines the list of plugins for each role
> 
> This patch is adding another directory to search for plugins to a context, maybe this list of directories could be defined per role.
> 
> It is just an idea (maybe a bad one), thatk would help to fix similar issues in the future on any role by applying changes to the role configuration, not the testing tool.

I don't think we need a per-role list of directories to search - it is the same list of directories for all roles.  That is - every role may use `tests/library` for test plugins - it's just that for now only certificate uses this.  I guess it could be an optimization to avoid looking into certain directories on a per-role basis, but IMO checking for such a per-role list is approximately as expensive as looking for a `tests/library` that may or may not exist.

report-modules-plugins.py
```diff
@@ -602,6 +602,7 @@ def process_role(role_path, is_real_role, ctx):
             ctx.found_role_name = get_role_name(role_path)
         ctx.enter_role(ctx.found_role_name, role_path)
         ctx.add_local_plugins(role_path, "library")
+        ctx.add_local_plugins(role_path, "tests/library")
```
richm, line 605: This should really be done in the `tests` context.  Consider this case:

Developer accidentally uses the plugin defined in `tests/library` in a `tasks/*.yml` file.  Then `report-modules-plugins.py` will pass, and the integration tests will pass, but the role will fail at runtime.

report-modules-plugins.py
```diff
@@ -569,7 +569,7 @@ def process_playbooks_path(playbooks_path, ctx):
         if os.path.isfile(itempath):
             process_ansible_file(itempath, ctx)
     # treat it like a role - look for vars/, tasks/, etc.
-    process_role(playbooks_path, False, ctx)
+    process_role(playbooks_path, True, ctx)
```
richm, line 572: That's it?  And this works?  If it doesn't work with big collections like community.general, that's ok, as long as it works with all roles in CI.  We can then worry about adding support for other collections later.

richm:
Can we merge this?

spetrosi:
I run it against all roles, it does work. It works also when I run it against community.general in `plugins` directory, but not in the parent directory. And changing True back to False doesn't help.

PR #311, opened 2023-10-23
manage role fixes

richm:
- do not fork empty repos
- do not release template role by default
- change wording to release instead of skip


spetrosi (commented):
Thanks for the fix, lgtm

PR #324, opened 2024-01-10
use ANSIBLE_JINJA2_NATIVE=1 for set_fact value: Yes

richm:
otherwise, doing
```
set_fact:
  var: "Yes"
```
will convert `var` to a boolean.  We want it to be a string.


richm:
See https://github.com/ansible/ansible/pull/43425

PR #327, opened 2024-01-17
fix: release_collection should use code matching tag

richm:
release_collection was not ensuring that the code checked out to
use to build the collection matched the tag/hash specified.

The code will now ensure:
* if --no-update was used, just ensure the given cur_ref is
  checked out, or main HEAD if cur_ref is None
* if cur_ref was set to a tag or commit hash, and there have been
  no updates to the role code since that tag/hash, the tag/hash will
  be checked out
* if there have been updates to the role code since the cur_ref
  * if there is a new tag, check out that tag
  * elif use_commit_hash is set, use main HEAD which is already checked
    out
  * else there have been role updates, but not yet tagged, and we
    are not using the commit hash - ensure cur_ref is checked out

### DIFF
--- a/release_collection.py
+++ b/release_collection.py
@@ -208,6 +208,19 @@ def comp_versions(cur_ref, new_ref):
             return -1  # assume new is "newer" than cur
 
 
+def ref_is_hash(ref):
+    """Is ref a git commit hash?"""
+    if not ref:
+        return False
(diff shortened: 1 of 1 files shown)


richm:
I want to get this merged before the next nightly build.  I've tested it locally and it works.

PR #354, opened 2024-10-04
manage_issues.py with ability create multiple issues and epics

richm:
This version of manage_issues.py allows creation of multiple issues and
epics at the same time, and links the created issues to the created
epic.  It also allows creation of issues from github issues and PRs,
similar to manage-issues.sh.  There is still plenty of missing
functionality of manage-issues.sh.

### DIFF
--- a/manage_issues.py
+++ b/manage_issues.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright: (c) 2020, Red Hat, Inc.
+# Copyright: (c) 2024, Red Hat, Inc.
 # SPDX-License-Identifier: MIT
-"""This script is used to manage Bugzilla and Jira issues
+"""This script is used to manage Jira issues
 List and perform various lifecycle tasks
 """
 
--- a/manage_issues.py
+++ b/manage_issues.py
@@ -1,0 +1,576 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2020, Red Hat, Inc.
+# SPDX-License-Identifier: MIT
+"""This script is used to manage Bugzilla and Jira issues
+List and perform various lifecycle tasks
+"""
+
+import click
+import json
+import logging
+import os
+import requests
+import signal
+import sys
+
+from jira import JIRA
+
+try:
+    import yaml
+except ImportError:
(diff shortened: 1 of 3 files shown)


richm:
The jira python api has improved quite a bit since I last tried to use it - I don't think we should use the ansible based api and manage-issues.sh going forward.

The reason I'm using `click` instead of `argparse` is that `click` allows you to use the same subcommand multiple times, but argparse does not - so with click we can do `manage_issues.py create_issue --summary '' ... create_issue ... create_issue ...`

manage_issues.py
```diff
@@ -0,0 +1,579 @@
...
+
+# these are all fields that can use the string format method
+# to construct the final value from variables derived from
+# previous issues
+TEMPLATE_FIELDS = ["description", "epic_name", "summary"]
+
+
+# this handles creating the jira issue, updating it,
+# doing the status transition, adding links, etc.
+def __create_issue(kwargs):
```
spetrosi, line 318: Maybe rename this to __manage_issue if this handles updating them too?

manage_issues.py
```diff
@@ -0,0 +1,579 @@
...
+    help="One or more role names",
+)
+@click.option(
+    "--debug",
+    is_flag=True,
+    help="Turn on debug logging.",
+)
+@click.option(
+    "--doc-text-type",
+    type=str,
```
spetrosi, line 503: Can script specify available types like for summary like in `type=click.Choice(["Bug", "Story", "Task", "Epic"], case_sensitive=False),`

spetrosi (commented):
Looks good, I'd appreciate some usage examples

manage_issues.py
richm: Right now it only handles creating new issues - note that the way that jira works, you cannot set all of the fields in the create operation - some fields, like story points, must be set after creation - that's why `__create_issue` calls `__update_issue` if there are fields which cannot be set in the create operation.

manage_issues.py
richm: yes

richm:
The way the `--base-issue` flag works is this - let's say there is already an opened issue for a bug or a feature - you want to create the tracking tasks for that - you don't need to create or update that issue, but you want to use the summary, labels, and github links from that issue for the tracking tasks, and for the epic

jharuda:
lgtm

PR #356, opened 2024-10-09
Add support for main.yaml in a role instead of only supporting main.yml

### LINKED ISSUE #355: lsr_role2collection.py does not support main.yaml files in roles
I was facing an issue when I tried to use the `lsr_role2collection.py` with a role, that uses a `main.yaml` file in `tasks` instead of a `main.yml`.

hhenkel:
This resolves #355 for me

### DIFF
--- a/lsr_role2collection.py
+++ b/lsr_role2collection.py
@@ -688,6 +688,21 @@ def file_replace(path, find, replace, file_patterns):
                 with open(filepath, "w", encoding="utf-8") as f:
                     f.write(s)
 
+def main_file_exists(src_path, role):
+    filenames = ["main.yml", "main.yaml"]
+    paths = [
+        os.path.join(src_path, "tasks"),
+        os.path.join(src_path, role, "tasks")
+    ]
+
+    for path in paths:
+        for filename in filenames:
+            logging.debug(f"Testing if  {os.path.join(path, filename)} is a file.")
+            if os.path.isfile(os.path.join(path, filename)):
+                logging.debug(f"{os.path.join(path, filename)} exists and is a file - setting src_path to {path}.")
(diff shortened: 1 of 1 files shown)


lsr_role2collection.py
```diff
@@ -688,6 +688,21 @@ def file_replace(path, find, replace, file_patterns):
...
+        os.path.join(src_path, "tasks"),
+        os.path.join(src_path, role, "tasks")
+    ]
+
+    for path in paths:
+        for filename in filenames:
+            logging.debug(f"Testing if  {os.path.join(path, filename)} is a file.")
+            if os.path.isfile(os.path.join(path, filename)):
+                logging.debug(f"{os.path.join(path, filename)} exists and is a file - setting src_path to {path}.")
+                src_path = os.path.join(path)
```
richm, line 703: Setting `src_path` here does not set `src_path` in the calling context.  Maybe have this function return `src_path` if the main YAML file exists, or raise an error if the file does not exist?

lsr_role2collection.py
```diff
@@ -1338,12 +1353,8 @@ def parse_extra_mapping(mapping_str, namespace, collection, role):
...
-    _tasks_main = src_path / "tasks/main.yml"
-    if not _tasks_main.exists():
-        src_path = src_path / role
-        _tasks_main = src_path / "tasks/main.yml"
 
-    if not _tasks_main.exists():
+    if not main_file_exists(src_path, role):
         logging.error(
             f"Neither {src_path} nor {src_path.parent} is a role top directory."
         )
```
richm, line 1360: replace the `if` and the `logging` with just
```python
    src_path = main_file_exists(src_path, role)
```

richm:
ping

Issue #357, opened 2024-10-09
release_collection.py does not support multiple remote repos


hhenkel:
As I'm currently trying to adapt your tooling to release our internal roles as a collection I'm facing an issue, if I have two or more repositories that have to be checked out from github. Within the `release_colelction.py`  the `lsr_role2collection.py` gets executed with the `--src-path` argument set to the tmp folder. The issue here is, that when you have multiple remote repositories, all of the repositories are checked out to the tmp folder and the `lsr_role2collection.py` picks up the wrong role.

```
INFO:root:The role [usergroups] is updated. Getting changelog entries from [0.0.0] to [0.0.1].
DEBUG:root:/home/hhenkel/venvs/ansible-navigator/bin/python3 lsr_role2collection.py --src-owner linux-system-roles --role usergroups --src-path /tmp/collection_cu_ddjk.lsr --dest-path /tmp/ansible-collection --na
mespace my_corp --collection linux_base --subrole-prefix private_usergroups_subrole_ --readme lsr_role2collection/collection_readme.md returned 1 stdout  stderr Traceback (most recent call last):
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 1824, in <module>
    sys.exit(role2collection())
             ^^^^^^^^^^^^^^^^^
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 1713, in role2collection
    lsr_copytree(extra, dest)
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 649, in lsr_copytree
    lsr_copytree(
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 649, in lsr_copytree
    lsr_copytree(
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 649, in lsr_copytree
    lsr_copytree(
  File "/home/hhenkel/git/github/linux-system-roles/auto-maintenance/lsr_role2collection.py", line 659, in lsr_copytree
    copy2(subsrc, subdest, follow_symlinks=(not symlinks))
  File "/usr/lib64/python3.12/shutil.py", line 475, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib64/python3.12/shutil.py", line 262, in copyfile
[...]

PR #358, opened 2024-10-09
Change source path to the actual role folder

hhenkel:
Prevent issues when checking out multiple remote roles.

Fix for #357

### DIFF
--- a/release_collection.py
+++ b/release_collection.py
@@ -409,7 +409,7 @@ def role_to_collection(
         "--role",
         rolename,
         "--src-path",
-        args.src_path,
+        os.path.join(args.src_path, rolename),
         "--dest-path",
         args.dest_path,
         "--namespace",


release_collection.py
```diff
@@ -409,7 +409,7 @@ def role_to_collection(
         "--role",
         rolename,
         "--src-path",
-        args.src_path,
+        os.path.join(args.src_path, rolename),
```
richm, line 412: `src_path` can be either the parent directory of the role, or the role directory.  Forcing it to be only the role directory will break some of our use cases.

hhenkel: @richm Okay - so what be an acceptable solution then? The only other thing that currently comes to my mind is, to checkout remote roles to individual tmp folders. As you can see in the provided error log in the issue #357 you code currently picks up the wrong folder on the second cloned repository.

richm: @hhenkel What repo(s) are `release_collection.py` checking out from github?  What does your `collection_release.yml` look like?

hhenkel: It looks like this:
```
usergroups:
  org: Github-My_corp-CHE
  repo: cioapir-my_corp.ansible-role-usergroups
  ref: 0.0.0
  sourcenum: 1
users:
  org: Github-My_corp-CHE
  repo: cioapir-my_corp.ansible-role-users
  ref: 0.0.0
  sourcenum: 2
sudo:
  org: Github-My_corp-CHE
  repo: cioapir-my_corp.ansible-role-sudo
  ref: 0.0.0
  sourcenum: 3
```

PR #361, opened 2024-11-11
check logs

richm:


spetrosi:
Could you add examples of running it to get logs from github and from the cold storage for history?

PR #368, opened 2025-02-14
handling testing farm logs

richm:
use `--tf-job-url https://api.dev.testing-farm.io/v0.1/requests/xxxxx`
to pass in the testing farm api url.  The job must already be
completed.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -235,9 +235,18 @@ def get_statuses(gh, org, repo, pr_num):
     return statuses
 
 
+# beaker
 SYSTEM_ROLE_LOG_RE = re.compile(
     r"/SYSTEM-ROLE-(?P<role>[a-z0-9_]+)_(?P<test_name>tests_[a-z0-9_]+"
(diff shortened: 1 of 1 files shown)


richm:
@spetrosi @xjezda00

check_logs.py
```diff
@@ -252,30 +261,47 @@ def get_errors_from_ansible_log(args, log_url):
...
+                    role,
+                    str(args.role),
+                    log_url,
+                )
+                return []
+            # test = match.group(2)  # unused for now
+            ansible_version = match.group("ansible_ver")
+        else:
+            # testing farm - https://...../data/$ROLENAME-$TESTNAME-ANSIBLE-$VER-$TFTESTNAME-$STATUS.log
+            match = SYSTEM_ROLE_TF_LOG_RE.search(log_url)
```
spetrosi, line 292: Note that David suggested removing ROLE_NAME completely to make it work like upstream_testsuite. So that when `SYSTEM_ROLES_ONLY_TESTS` is not provided than we test all roles.
But that's in planning stage, we can revise this later.

spetrosi (approved):

PR #371, opened 2025-03-31
show last line of beaker log

richm:


spetrosi:
What is handy info on the last line?

richm:
> What is handy info on the last line?

For example - checking the logs of Upstream-testsuite - when doing all of the tests for a role, you don't get any PASS or FAIL for individual tests until all tests are completed - for a role with a lot of tests like storage, it looks like the test is hung - with this last line addition, you can see that the test is proceeding

PR #376, opened 2025-05-06
callback plugin to collect task errors

richm:
We will add this to all of our test runners - tox runqemu, testing farm, etc. - this will create a file `.json` corresponding to the log which will list the errors in the tasks in an easy to digest format - no more log parsing

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import argparse
-import ast
 import copy
 import csv
 import datetime
@@ -22,7 +21,7 @@ try:
     from ghapi.all import GhApi
     from ghapi.page import paged as gh_paged
 except ModuleNotFoundError:
-    logging.warning("no ghapi library")
+    pass
 from bs4 import BeautifulSoup
 from http.client import HTTPConnection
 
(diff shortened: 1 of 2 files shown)


richm:
@spetrosi The way we will use this is to create a `callback_plugins/` sub-directory under the `tests/` directory with the file lsr_report_errors.py.  This will write the error report in the regular output log, then check_logs.py can easily read it out.

The next step will be to change our test frameworks to create the `callback_plugins/` sub-directory and copy the lsr_report_errors.py script there.

richm:
@martinpitt @spetrosi this is what it looks like when the loop items are de-duplicated:

```
    {
        "ansible_version": "2.16.14",
        "delta": "0:00:00.002142",
        "end_time": "2025-05-13 12:07:41.278205",
        "host": "managed-node3",
        "loop_item": 3,
        "loop_label": "<same as loop_item>",
        "loop_var": "item",
        "message": "non-zero return code",
        "parents": [],
        "rc": 1,
        "start_time": "2025-05-13 12:07:41.276063",
        "task_name": "Try loop failure",
        "task_path": "/home/rmeggins/ansible_sandbox/lsr_error_test.yml:10"
    },
```
In this case, the message is different, but the item and label are the same.

richm:
Here's what it looks like when the message, item, and label are all the same:
```
    {
        "ansible_version": "2.16.14",
        "delta": "0:00:00.001884",
        "end_time": "2025-05-13 12:31:52.903548",
        "host": "managed-node3",
        "loop_item": "<same as message>",
        "loop_label": "<same as message>",
        "loop_var": "item",
        "message": "non-zero return code",
        "parents": [],
        "rc": 1,
        "start_time": "2025-05-13 12:31:52.901664",
        "task_name": "Try loop failure",
        "task_path": "/home/rmeggins/ansible_sandbox/lsr_error_test.yml:10"
    },
```

check_logs.py
```diff
@@ -22,7 +21,7 @@
     from ghapi.all import GhApi
     from ghapi.page import paged as gh_paged
 except ModuleNotFoundError:
-    logging.warning("no ghapi library")
+    pass
```
martinpitt, line 24: Was this change intentional?

martinpitt (commented):
I didn't review this too deeply (sorry, got too absorbed into firewall today), but I don't see any obvious blocker. Just two questions, but we can put this into practice and then fix it up as necessary -- this is neither going into products nor is it even enabled yet, and the enabling will happen in PRs. So should be quite safe.

check_logs.py
richm: yes - I got tired of seeing this message - I originally thought it would be helpful, but it isn't

martinpitt (approved):
Thanks!

PR #379, opened 2025-05-15
format log output for github actions

richm:
Add `--github-action-format` argument to tell check_logs to format
errors in a github action friendly output format.

Looks like this:

```
::group::tests_auth_and_security.yml.log Ensure container images are present
Role: [podman] Ansible Version: [2.17.11]
Task Path: /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/create_update_kube_spec.yml:39
Parents:
    /path/to/linux-system-roles/podman/tests/tests_auth_and_security.yml:54
    /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/main.yml:210
    /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/handle_kube_spec.yml:135
Log url: /path/to/linux-system-roles/podman/tests/tests_auth_and_security.yml.log
Detail:
the output has been hidden due to the fact that 'no_log: true' was specified for this result
::endgroup::
::group::tests_auth_and_security.yml.log Ensure container images are present
Role: [podman] Ansible Version: [2.17.11]
Task Path: /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/create_update_kube_spec.yml:39
Parents:
    /path/to/linux-system-roles/podman/tests/tests_auth_and_security.yml:80
    /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/main.yml:210
    /path/to/linux-system-roles/podman/tests/roles/linux-system-roles.podman/tasks/handle_kube_spec.yml:135
Log url: /path/to/linux-system-roles/podman/tests/tests_auth_and_security.yml.log
Detail:
the output has been hidden due to the fact that 'no_log: true' was specified for this result
::endgroup::
```

## Summary by Sourcery

Implement GitHub Actions formatting in check_logs by adding a CLI option that wraps each error in ::group:: blocks with comprehensive task information.

New Features:
- Add --github-action-format/--gh-format flag to output errors in a GitHub Actions friendly format

Enhancements:
- Format error output into collapsible GitHub Actions groups showing role, Ansible version, task path, parent tasks, log URL, and detail

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -962,6 +962,10 @@ def print_testing_farm_result(args, result):
     print("")
 
 
+def sanitize_for_actions(text):
+    return re.sub(r"^::", " ::", text, flags=re.M)
+
+
 def print_ansible_errors(args, errors):
     if not errors:
         print("No errors found")
@@ -1031,6 +1035,23 @@ def print_ansible_errors(args, errors):
         sh.values_update(
             worksheet.title, {"valueInputOption": "USER_ENTERED"}, {"values": values}
         )
+    if args.github_action_format:
+        for error in errors:
+            print(f"::group::{os.path.basename(error['Url'])} {error['Task']}")
+            print(
+                f"Role: [{error['Role']}] Ansible Version: [{error['Ansible Version']}]"
+            )
+            print(f"Task Path: {error['Task Path']}")
+            parents = error.get("Parents")
+            if parents:
+                print("Parents:")
+                for parent in parents:
+                    print(f"    {parent}")
+            print(f"Log url: {error['Url']}")
+            print("Detail:")
+            detail = sanitize_for_actions(error["Detail"])
+            print(detail)
+            print("::endgroup::")
 
 
 def parse_arguments():
@@ -1168,6 +1189,13 @@ def parse_arguments():
         action="append",
         help="url of testing farm job api e.g. https://api.dev.testing-farm.io/v0.1/requests/xxxxx",
     )
+    parser.add_argument(
+        "--github-action-format",
+        "--gh-format",
+        default=False,
+        action="store_true",
+        help="Write errors in GitHub Actions-friendly format",
+    )
 
     args = parser.parse_args()
     return args
@@ -1182,7 +1210,7 @@ def main():
     elif args.verbose > 0:
         logging.getLogger().setLevel(logging.INFO)
 
-    if args.csv_errors or args.gspread:
+    if args.csv_errors or args.gspread or args.github_action_format:
         args.gather_errors = True
 
     if args.testing_farm_job_url:


richm:
This will be used in https://github.com/linux-system-roles/.github/blob/main/playbooks/templates/.github/workflows/qemu-kvm-integration-tests.yml#L186 to print the errors

check_logs.py
```diff
@@ -1031,6 +1031,22 @@ def print_ansible_errors(args, errors):
...
+            )
+            print(f"Task Path: {error['Task Path']}")
+            parents = error.get("Parents")
+            if parents:
+                print("Parents:")
+                for parent in parents:
+                    print(f"    {parent}")
+            print(f"Log url: {error['Url']}")
+            print(f"Detail:")
+            print(f"{error['Detail']}")
```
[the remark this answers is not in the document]
martinpitt: @richm That's the point, right? to add grouping and such.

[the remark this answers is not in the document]
martinpitt: Well, hmm -- the groupign should happen explicitly, any `::command` inside of `Detail` should be escaped indeed (not sure how to do that -- but I suppose it doesn't matter much, `s/::/: :/` should already do it.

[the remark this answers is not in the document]
richm: Right - I've been looking to see if there is something like a way to mark a block of text as "raw - do not process" but there does not appear to be anything - while I think the chances of having `::` at the beginning of a line is very slim, it is non-zero. so I'll implement something like that.

check_logs.py
```diff
@@ -1031,6 +1031,22 @@ def print_ansible_errors(args, errors):
         sh.values_update(
             worksheet.title, {"valueInputOption": "USER_ENTERED"}, {"values": values}
         )
+    if args.github_action_format:
+        for error in errors:
```
martinpitt, line 1035: I'm a bit confused by this function -- It's called `print_ansible_errors`, but it creates a spreadsheet. With `--gh-format` should still actually do that, or is it more of an if/else thing? To my knowledge we don't put any spread sheet file into the test artifact.

richm: The output format options listed in `print_ansible_errors` are not mutually exclusive - you could output all of CSV, a google spreadsheet, and github action format at the same time.  That's probably unlikely.

richm: Note that `print_ansible_errors` will not output a google spreadsheet unless you pass in the `--gspread` and other related arguments.

martinpitt: Ack, thanks!

check_logs.py
```diff
@@ -962,6 +962,16 @@ def print_testing_farm_result(args, result):
...
 
 
+def sanitize_for_actions(text):
+    safe_lines = []
+    for line in text.splitlines():
+        # if it starts with :: we prepend a space so it can’t be parsed as a command
+        if line.startswith("::"):
+            line = " " + line
+        safe_lines.append(line)
+    return "\n".join(safe_lines)
```
martinpitt, line 972: Not sure if that's actually faster than `re.sub(r"^::", " ::", flags=re.M)`, but re would certainly be simpler. But bikeshedding/matter of taste.

check_logs.py
```diff
@@ -1031,6 +1041,23 @@ def print_ansible_errors(args, errors):
...
+            print(f"Task Path: {error['Task Path']}")
+            parents = error.get("Parents")
+            if parents:
+                print("Parents:")
+                for parent in parents:
+                    print(f"    {parent}")
+            print(f"Log url: {error['Url']}")
+            print("Detail:")
+            detail = sanitize_for_actions(error["Detail"])
+            print(f"{detail}")
```
martinpitt, line 1059: Unnecessary f-string wrapping.

martinpitt (approved):
Thanks! Some bikeshedding, and if you want to go for another round I'm happy to re-review (I'm EOD, so spotty, but can have another look in a bit)

martinpitt (approved):
Cheers!

PR #380, opened 2025-05-15
allow running without bs4; improve other missing library handling

richm:
We do not need bs4 unless parsing html or xml, which is not needed
for lsr error log processing.

Improve the exception handling and reporting for other missing libraries.

## Summary by Sourcery

Allow running the script without bs4 and standardize missing dependency handling with unified exception clauses and debug-level logging.

Enhancements:
- Make bs4 optional by wrapping its import in a try/except and logging a debug message if missing.
- Catch both ImportError and ModuleNotFoundError for ghapi and gspread imports and lower their missing-library logs to debug-level.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -20,18 +20,20 @@ import sys
 try:
     from ghapi.all import GhApi
     from ghapi.page import paged as gh_paged
-except ModuleNotFoundError:
-    pass
-from bs4 import BeautifulSoup
+except ImportError:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Haha @ "professional message" -- let's have some fun here, this is CI only. :grin: But the ImportError simplification was good. In general I find sourcery's feedback a lot more useful and palatable in system-roles than in cockpit.

PR #381, opened 2025-05-16
record task stdout, stderr; skip tasks that ignore error

richm:
Record the task stdout and stderr

Do not record errors for tasks that use `ignore_errors: true`
or `failed_when: false`.

Add a few other fields to github action output.

Omit fields that are empty.

## Summary by Sourcery

Enable richer error reporting by capturing additional task metadata (stdout, stderr, RC, timings, host) and skip tasks that explicitly ignore failures, while refining GitHub Action output to include only non-empty fields and renaming 'Url' to 'Log Url'.

New Features:
- Capture and record task stdout, stderr, return code, start/end times, and host metadata for failed tasks

Bug Fixes:
- Skip logging errors for tasks with ignore_errors: true or failed_when: false

Enhancements:
- Rename 'Url' field to 'Log Url' and omit empty fields in GitHub Action output
- Refactor error parsing and printing to loop over optional fields for conciseness

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -801,13 +801,25 @@ def parse_lsr_error_log(args, log_url, extra_fields={}):
                     "Ansible Version": error_item["ansible_version"],
                     "Task": error_item["task_name"],
                     "Task Path": error_item["task_path"],
-                    "Parents": error_item["parents"],
-                    "Url": log_url,
+                    "Log Url": log_url,
                     "Detail": error_item["message"],
                 }
             )
             if role:
                 error["Role"] = role
+            optional_fields = [
+                ("Parents", "parents"),
+                ("Stdout", "stdout"),
+                ("Stderr", "stderr"),
+                ("RC", "rc"),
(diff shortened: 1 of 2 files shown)


check_logs.py
```diff
@@ -801,13 +801,25 @@ def parse_lsr_error_log(args, log_url, extra_fields={}):
...
+                ("Stdout", "stdout"),
+                ("Stderr", "stderr"),
+                ("RC", "rc"),
+                ("Start", "start_time"),
+                ("End", "end_time"),
+                ("Host", "host"),
+            ]
+            for title, field in optional_fields:
+                value = error_item.get(field)
+                if value or value == 0:
```
spetrosi, line 821: For what case you do `value == 0:` ?

richm: the result code of a command or shell - `rc` - is an `int` and can have a `0` value, which we want to keep and show - it would be helpful in debugging to know that a task reported an error, but the `rc` is `0`

PR #425, opened 2026-06-10
Add information about collection requirement older versions

### LINKED ISSUE #137: Cannot install community.general because fedora.linux_system_roles has >=6.6.0,<12.0.0 set
I want to install the `logrotate` module, which was introduced in 12.5.0: https://docs.ansible.com/projects/ansible/latest/collections/community/general/logrotate_module.html

But it's not included in my environment, because the dependency map cannot move beyond version 11 because of requirement dependency operators from the Linux system roles.

```sh
[ERROR]: Failed to resolve the requested dependencies map. Could not satisfy the following requirements:
* community.general:12.6.1 (direct request)
* community.general:>=6.6.0,<12.0.0 (dependency of fedora.linux_system_roles:1.125.2)
* community.general:>=6.6.0,<12.0.0 (dependency of fedora.linux_system_roles:1.125.2)
```

Can we bump this version? Version 13.0.1 is already the latest version.

spetrosi:
In collection-requirements.yml in the built fedora.linux_system_roles collection, add comment that users can use latest collections if they don't need to support EL7.

Fixes https://github.com/linux-system-roles/linux-system-roles.github.io/issues/137

Related Jira: RHELMISC-34455

## Summary by CodeRabbit

* **Documentation**
  * Enhanced generated requirements guidance with explanatory comments clarifying EL7/collection dependency version limitations, so users receive clearer compatibility and installation guidance when inspecting requirement manifests.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -765,7 +765,9 @@ def get_beaker_job_info(args, job):
             xml_data = open(job).read()
         bs = BeautifulSoup(xml_data, "xml")
     else:  # assume it is a job number like J:1213552
-        result = subprocess.run(["bkr", "job-results", job], capture_output=True, text=True, check=True)
+        result = subprocess.run(
+            ["bkr", "job-results", job], capture_output=True, text=True, check=True
+        )
         bs = BeautifulSoup(result.stdout, "xml")
     data = {}
     data["job"] = job
--- a/release_collection.py
+++ b/release_collection.py
@@ -513,7 +513,15 @@ def create_collection_extra_files(args, coll_dir, galaxy=None):
             ee_dependencies.append("system: bindep.txt")
         if galaxy["dependencies"]:
             with open(collection_req_yml_dest, "w") as crf:
-                crf.write("---\ncollections:\n")
+                crf.write(
+                    "---\n"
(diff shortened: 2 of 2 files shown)


spetrosi:
Hi @richm, you think that's enough?

I also wanted to change this in each role's `collection-requirements.yml` via .github, but most probably this issue happens only for collections, not individual roles. So I guess this is fine.

The error reported in https://github.com/linux-system-roles/linux-system-roles.github.io/issues/137 happens when building execution environments, the fix that you suggested - force installing required version if managing EL7 is not required - should work fine.

richm (approved):

richm:
> Hi @richm, you think that's enough?
> 
> I also wanted to change this in each role's `collection-requirements.yml` via .github, but most probably this issue happens only for collections, not individual roles. So I guess this is fine.

It could happen for individual roles.  But the vast majority of cases will be the collection, so let's fix this and worry about roles if someone reports it.
 
> The error reported in [linux-system-roles/linux-system-roles.github.io#137](https://github.com/linux-system-roles/linux-system-roles.github.io/issues/137) happens when building execution environments, the fix that you suggested - force installing required version if managing EL7 is not required - should work fine.

+1

richm:
Looks like you'll need to rebase

PR #434, opened 2026-08-12
ci: use commit hash for github action, add persist-credentials false [citest_skip]

richm:
The latest security guidance is to use the full commit hash, which is immutable,
instead of a tag or version, which can be mutable, for the reference to a version
of a github action.  There are known attacks which inserted unauthorized code
in a version tag and moved the tag.  This prevents this sort of attack, at the
cost of more maintenance burden, but dependabot will largely take care of this
for us.

## Summary by CodeRabbit

* **Chores**
  * Pinned automation workflow actions to specific commits for more reliable and secure builds.
  * Updated publishing, pull request, and testing workflows while preserving their documented action versions.


richm:
@spetrosi can we get rid of the pull-request.yml workflow?  It has never been run - I don't remember what it was used for - it has some problems that need to be fixed if we decide to keep it.

PR #437, opened 2026-09-10
publish_collection: prune old collection versions from Galaxy

spetrosi:
Add remove_old_galaxy_versions.py to keep only the newest N versions (default 20) of a collection on Ansible Galaxy via the v3 REST API, and wire it into the publish workflow to run after publishing.

## Summary by CodeRabbit

* **Chores**
  * Automated publishing now retains only the 20 newest collection versions in Ansible Galaxy.
  * Added support for listing, sorting, and removing older published versions.
  * Added a dry-run option and safeguards for missing credentials or deletion errors.

### DIFF
--- a/.github/workflows/publish_collection.yml
+++ b/.github/workflows/publish_collection.yml
@@ -77,6 +77,7 @@ jobs:
           echo "updated=1" >> $GITHUB_OUTPUT
 
       - name: Keep latest 20 versions in Galaxy
+        if: ${{ steps.build.outputs.updated == 1 }}
         shell: bash
         run: |
           set -euxo pipefail
--- a/.github/workflows/publish_collection.yml
+++ b/.github/workflows/publish_collection.yml
@@ -76,6 +76,14 @@ jobs:
           echo "tagname=$NEWVER" >> $GITHUB_OUTPUT
           echo "updated=1" >> $GITHUB_OUTPUT
 
+      - name: Keep latest 20 versions in Galaxy
+        shell: bash
+        run: |
+          set -euxo pipefail
+          python3 ./remove_old_galaxy_versions.py --keep 20
+        env:
+          GALAXY_API_KEY: ${{ secrets.GALAXY_API_KEY }}
+
       - name: Create tag
         if: ${{ steps.build.outputs.updated == 1 }}
         uses: mathieudutour/github-tag-action@a22cf08638b34d5badda920f9daf6e72c477b07b # v6.2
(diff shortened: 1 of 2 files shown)


spetrosi:
I verified in a dry-run that correct collections are removed and the latest 20 collections kept.
The next run would remove 261 collections and keep newest 20.
I am going to use this script for microsoft.sql and infra.leapp collections in galaxy too. Namespace and name are configurable.

@richm do you think keeping 20 versions is good enough? 
If we keep only the newest 20 versions of fedora.linux_system_roles:

- Oldest kept: 1.124.0, published 2026-05-07 (~4 months ago as of today, 2026-09-10).
- Newest kept: 2.4.2, published 2026-08-24.

So keeping 20 gives roughly a 4-month retention window.

spetrosi:
Closing this — we've decided not to trim old versions of fedora.linux_system_roles on Galaxy. After investigation, removing versions would do more harm than good and wouldn't actually fix the underlying problem.

Why we're not trimming:

* Users pin specific versions. Someone may need e.g. fedora.linux_system_roles: 1.84.3 to support an older system/component and cannot move to a newer release. Deleting old versions would break ansible-galaxy collection install ...:<version> for them. We already found real public examples of users pinning old versions:
  https://github.com/genevieve-me/home-services/blob/main/requirements.yml#L5-L6

  https://github.com/wzzrd/ansible-dev-on-apple/blob/main/requirements.yml#L4-L5

* We can't measure who relies on old versions. Galaxy doesn't expose per-version download counts — only a collection-wide total — so there's no safe way to know which versions are in active use before deleting them.

* Other large collections don't trim. community.general, for example, keeps all versions and doesn't trim at all. Smaller collections simply don't have enough versions for it to matter.

* Neither trimming strategy is sound. By-time removal misbehaves for collections that publish rarely (a whole year could pass with a single release); by-count is somewhat better but still deletes versions people may have pinned.

The original problem was different. This started because infra.leapp depended on fedora.linux_system_roles:*, causing ansible-galaxy to churn through version metadata. That's not really a system-roles problem, and it has since been addressed on that side. The remaining slowness/timeouts (e.g. ansible-galaxy collection install fedora.linux_system_roles timing out on RHEL 7, ~6m20s; ~2m30s on RHEL 8; ~3m on RHEL 9) come from how ansible-galaxy processes the dependency map and fetches v-side limitation, not somethingwe fix by deleting our publishedning an exact version (==2.4.2)doesn't help on EL7, since that  out just processing the
requirements/dependency map.

In short: trimming old versions would risk breaking real users' pinned installs while not addressing the actual (Ansible-side) cause of slow installs. Closing as won't-do.

